PR #42, opened 2019-11-12
41 missing color preambles

yaleLeeNGA:
Should cover all missing preambles. Needs to me pulled soon to make sure new commits include this new way of printing log, and avoid behavior gap.

### DIFF
--- a/src/Applications/NodeGossiper.py
+++ b/src/Applications/NodeGossiper.py
@@ -44,12 +44,12 @@
 ###############################################################################
 NodeGossiper_module_aliases = {}
 for m in [
-    "os",
-    "subprocess",
+    "bcolors",
     "getopt",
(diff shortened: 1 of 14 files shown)


src/Applications/NodeGossiper.py
```diff
@@ -44,12 +44,12 @@
...
 NodeGossiper_module_aliases = {}
 for m in [
-    "os",
-    "subprocess",
+    "bcolors",
     "getopt",
     "math",
+    "os",
+    "subprocess",
     "sys",
```
ppebay: Good practice to use alphabetical order!

yaleLeeNGA: Thanks :) Trying to implement as I edit files

Issue #53, opened 2020-06-23
Add support for subphases from vt applications


PhilMiller:
See https://github.com/DARMA-tasking/vt/issues/708 for details of why this is desired, and the code that produces the stats files containing per-subphase timing data.

lifflander:
Currently, the LB stats file outputs a format as so for each phase.

```
<iter/phase>, <object-id>, <time-in-seconds>
<iter/phase>, <object-id1-to/recv>, <object-id2-from/send>, <num-bytes>, <comm-type={1..6}>
```

In the new format, introduced by https://github.com/DARMA-tasking/vt/issues/708, we have added sub-phases to the format for the line with `<time-in-seconds>`. In particular, this commit https://github.com/DARMA-tasking/vt/commit/996d32ac465c0091e1acc208f2d2c2ad445afeff is where the format what changed. The new format looks like this:

```
<iter/phase>, <object-id>, <time-in-seconds> <#-of-subphases> '[' [<subphase-time-1>] ... [<subphase-time-N>] ']'
<iter/phase>, <object-id1-to/recv>, <object-id2-from/send>, <num-bytes>, <comm-type={1..6}>
```

We want to adjust the reader in the LBAF to read this format and process the new sub-phase data. Let's start with just getting it to read the data, then create the data structures in the simulator for processing the sub-phases.

- For changing the reader, look at this file: https://github.com/DARMA-tasking/LB-analysis-framework/blob/develop/src/IO/lbsLoadReaderVT.py
- You will also want to adjust the writer so it's able to write this back out: https://github.com/DARMA-tasking/LB-analysis-framework/blob/develop/src/IO/lbsLoadWriterVT.py

Let me know if you have questions. @nlslatt is running the app with the stats output so you have a valid input deck for the simulator to work with.

bradybray:
@lifflander I cloned the repo and added a new branch for myself. I'll file a pull request once finished for review.

lifflander:
@bradybray It turns out that we don't have the full application run with the new stats. That change didn't get merged onto the release branch that EMPIRE (the app) is using. Instead, I will attach some files of a benchmark locally so you have something to work with.

lifflander:
Here's the link on building VT: https://darma-tasking.github.io/docs/html/vt-build.html

It doesn't mention it there, but if you don't want to rebuild the base image yourself, do a `docker-compose pull ubuntu-cpp-interactive` or `docker-compose pull ubuntu-cpp`

[...]

PR #54, opened 2020-06-27
53 subphase support

bradybray:
Merging in development branch for review. Added subphase support. Unsure how to execute scripts to test against live file but used test file provided by @lifflander to develop process.


lifflander:
@bradybray Can you prefix your commit messages with the GitHub issue number for reference? On the VT repository we have automatic checks for this, but not on this on:

Example commit message: "#53: adding processing for subphases"

This helps us refer to the work done by each commit via the issue

src/IO/lbsLoadReaderVT.py
```diff
@@ -140,14 +140,19 @@ def read(self, node_id, time_step=-1, comm=False):
                 n_entries = len(row)
 
                 # Handle three-entry case that corresponds to an object load
```
PhilMiller, line 142: This comment is no longer accurate

src/IO/lbsLoadReaderVT.py
```diff
@@ -140,14 +140,19 @@ def read(self, node_id, time_step=-1, comm=False):
...
+                    #   <time_step/phase>, <object-id>, <time>, <#-subphases> '[' 
+                    #   [<subphase-time-1>] ... [<subphase-time-N>] ']'
                     # Converting these into integers and floats before using them or
                     # inserting the values in the dictionary
                     try:
                         phase, o_id = map(int, row[:2])
                         time = float(row[2])
+                        Nsubphases = int(row[3])
+                        # Trim first and last brackets
+                        row[4] = float(row[4][1:])
```
PhilMiller, line 154: Is this leaving all the subphase times between the first and last as unparsed text?

src/IO/lbsLoadReaderVT.py
```diff
@@ -169,13 +174,22 @@ def read(self, node_id, time_step=-1, comm=False):
...
                                 o_id,
-                                time))
+                                time,
+                                Nsubphases,
+                                end=''))
+                                
+                            for i in range(0,Nsubphases):
+                                print("subphase-time-" + str(i) + "= {}".format(
+                                    row[4 + i]
+                                end='')
```
PhilMiller, line 187: If we had a parsing error, then we can't trust that the value of `Nsubphases` accurately represents the content of the row we failed to parse.

src/IO/lbsLoadReaderVT.py
```diff
@@ -169,13 +174,22 @@ def read(self, node_id, time_step=-1, comm=False):
...
                                 + "[LoadReaderVT] "
                                 + bcolors.END
-                                + "iteration = {}, object id = {}, time = {}".format(
+                                + "iteration = {}, object id = {}, time = {}, subphases = {}".format(
                                 phase,
                                 o_id,
-                                time))
+                                time,
+                                Nsubphases,
+                                end=''))
```
PhilMiller, line 182: Not the fault of this PR, but this entire print is questionable - we failed at parsing and hence assigning these variables, so try printing the variables that we didn't parse? The print should be of the (ideally unmodified) `row`

src/IO/lbsLoadReaderVT.py
bradybray: A parsing error? As in the CSV reader fails?

src/IO/lbsLoadReaderVT.py
bradybray: Yes, that would appear to be the case. Given there is a number of sub phases there can be an additional loop that converts the strings to their respective types (floats I would assume).

bradybray:
> 
> 
> Can you please test with the new dataset `data/vt_example_lb_stats_subphases/stats.0.out` added by @lifflander and output result for comparison? Given the fact that the LB algorithm is unchanged (only considers phases), the result should be identical to that obtained without sub-phases. Or @PhilMiller @lifflander are we understanding this issue has having to perform LB after each subphase too?

Yes. I'll pull in the modules and try to run them against the included stat files.

src/IO/lbsLoadReaderVT.py
PhilMiller: I think you could do the same as seen a couple lines above, with `map(float, row[first:last])`. Given that, it would be nice if trimming out the brackets were applied separately from parsing the values.

I.e.
```
row[4] = row[4][1:]
row[-1] = row[-1][:-1]
row[4:] = map(float, row[4:])
```

src/IO/lbsLoadReaderVT.py
PhilMiller: The parsing errors that could be caught by this `except` clause would be when the values within the row weren't suitable for conversion to `int` or `float`. If that happens, some of the variable assignments would have failed, and so our exception handling would raise new exceptions on accesses to undefined variables. Worse, if the failure is not on the first row, by Python's variable scoping rules, those variables *will* be initialized, with the values from previous loop iterations, and the error message will be horribly confused.

src/IO/lbsLoadReaderVT.py
PhilMiller: Actually, better to not modify the row any more, and assign a separate variable

```
subphase_times = row[4:]
assert len(subphase_times) == Nsubphases
subphase_times[0] = subphase_times[0][1:]
subphase_times[1] = subphase_times[-1][:-1]
subphase_times = map(float, subphase_times)
```

src/IO/lbsLoadReaderVT.py
```diff
@@ -169,13 +175,22 @@ def read(self, node_id, time_step=-1, comm=False):
...
+                            for i in range(0,Nsubphases):
+                                print("subphase-time-" + str(i) + "= {}".format(
+                                    row[4 + i],
+                                end=''))
+                                
+                            print()
 
                 # Handle four-entry case that corresponds to a communication weight
-                elif n_entries == 5:
+                elif '[' not in row[4]:
```
PhilMiller, line 193: The `if` and `elif` cover all possibilities. This `elif` should keep itself more narrowly tailored to the case it actually handles, so that the `else` still gets run for lines that aren't ever going to be parsed correctly

bradybray: My latest commit addresses this issue.

bradybray: @PhilMiller 
https://github.com/DARMA-tasking/LB-analysis-framework/blob/30477157960faf905b351d385652fe17ae424c28/src/IO/lbsLoadReaderVT.py#L194

lifflander (approved):
Looks like this is ready to me

PR #73, opened 2021-02-18
72 fix incorrect empirical cmf computation under new criterion

ppebay:
The CMF problem is now fixed and it seems that the performance of LBAF was even slightly improved. If approved that change should go into GossipLB

### DIFF
--- a/src/Model/lbsProcessor.py
+++ b/src/Model/lbsProcessor.py
@@ -41,7 +41,7 @@
 ###############################################################################
 #@HEADER
 #
-########################################################################
+
 import math
 import random as rnd
 import sys
@@ -248,6 +248,7 @@ class Processor:
         # Update last received message index
         self.round_last_received = msg.get_round()
 
+
     def compute_cmf_underloads(self, l_ave, pmf_type=0):
         """Compute CMF of underloads given an average load
         """
@@ -257,14 +258,18 @@ class Processor:
 
         # Distinguish between different PMF types
         if not pmf_type:
-            # Initialize ancillary values
+            # Determine whether one underloaded is actually overloaded
+            loads = self.known_underloads.values()
+            l_max = max(loads)
+            factor = l_max if l_max > l_ave else l_ave
+                
+            # Initialize CMF value
             sum_p = 0.
-            inv_l_ave = 1. / l_ave
-
-            # Iterate over all underloads
-            for l in self.known_underloads.values():
+            
+            # Iterate over all loads known to be underloaded
+            for l in loads:
                 # Update CMF
-                sum_p += 1. - inv_l_ave * l
+                sum_p += 1. - factor * l
 
                 # Assign CMF for current underloaded processor
                 cmf.append(sum_p)
--- a/src/Model/lbsProcessor.py
+++ b/src/Model/lbsProcessor.py
@@ -267,7 +267,7 @@ class Processor:
         if not pmf_type:
             # Determine whether one underloaded is actually overloaded
             l_max = max(loads)
-            p_fac /= (l_max if l_max > l_ave else 1 / l_ave)
+            p_fac /= (l_max if l_max > l_ave else l_ave)
 
         elif pmf_type == 1:
             # User sender load
--- a/src/Model/lbsProcessor.py
+++ b/src/Model/lbsProcessor.py
@@ -251,7 +251,7 @@ class Processor:
 
     def compute_cmf_underloads(self, l_ave, pmf_type=0):
         """Compute CMF of underloads given an average load
-           tyoe 0: improved Gossip approac=h
+           tyoe 0: improved Gossip approach
            type 1: NS variant based on sender load
         """
 
--- a/src/Model/lbsProcessor.py
+++ b/src/Model/lbsProcessor.py
@@ -261,13 +261,14 @@ class Processor:
             # Determine whether one underloaded is actually overloaded
             loads = self.known_underloads.values()
             l_max = max(loads)
-            factor = l_max if l_max > l_ave else l_ave
-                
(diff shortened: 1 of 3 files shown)


src/Model/lbsProcessor.py
```diff
@@ -257,14 +258,18 @@ def compute_cmf_underloads(self, l_ave, pmf_type=0):
 
         # Distinguish between different PMF types
         if not pmf_type:
-            # Initialize ancillary values
+            # Determine whether one underloaded is actually overloaded
+            loads = self.known_underloads.values()
+            l_max = max(loads)
+            factor = l_max if l_max > l_ave else l_ave
```
nlslatt, line 264: Computing `factor` this way will lead to a zero contribution to the CMF for the highest loaded processor being considered, right?  If there is only one that becomes overloaded, we won't be able to select it for more transfers.  It will definitely change the CMF values for the other processors, however, making it slightly more likely that we choose an underloaded with higher load vs lower load.

nlslatt: We could potentially calculate `factor` using the load of the truly overloaded processor that is computing the CMF.  Due to making it more likely to choose less severely underloaded processors, however, I'm not sure if it would have the desired effect.

ppebay: @nlslatt thanks for the review & comments

By **the** truly OL processor, do you mean the one that is the most overloaded (amongst all those that were previously underloaded prior to a first round of object transfers) across all ranks? If so, making its load known to all in order to compute the CMF with the same normalizing factor would require some global communication as not all ranks would necessarily about it otherwise. 
But indeed in the approach above the normalization is not done in the same way across all ranks (those who are not aware of "overloaded-underloaded" targets simply divide by the average load).

nlslatt: @ppebay No, by the truly overloaded processor, I meant the processor computing the CMF, i.e., the one that is considering offloading work to a less-loaded processor.  Because the modified grapevine criterion involves not letting the selected processor become more overloaded than the one offloading the work, we know that the offloading processor will not make any processor included in the CMF calculation become more overloaded than itself.  Instead of taking `factor` from the maximum load included in the CMF calculation, we could take it from the (necessarily greater) load of the offloading processor.  We would still avoid negatives in the CMF, and would have non-zero CMF contributions from all included processors.  Whether doing so would give us extra wiggle room or lead us down the wrong path isn't immediately clear to me.

ppebay: @nlslatt 
Regarding your first comment about `factor` I agree with what you are describing, but 2 things to be considered here:
- it is similar to what would happen for a processor whose load is exactly at the average (and if no other rank has become overloaded): it would not get chosen. We could alleviate that by multiplying the normalization factor by 0.9 or 0.99, which I have tried but does not yield noticeable effects either way;
- only **the** most overloaded rank, amongst all known "underloaded" processors known to the sender, will have a 0 contribution to the CMF and therefore not be selected **at that iteration**, so
  -  in subsequent iterations it might no longer be the most overloaded rank in the list, and therefore become "actionable" again ;
  - and if several "known underloaded" have become overloaded, only the ones with maximal load will have that 0 contribution. Other, "less overloaded" ranks would have a non-0 contribution (as their L_max:L ratios would be smaller than 1) and therefore would remain "selectable".

ppebay: @nlslatt understood now regarding the max load to be used for the CMF. That's an interesting angle. I shall experiment with it a little. Pushing a change to the code now BTW

ppebay: @nlslatt Nicole I looked into histograms given by your proposed variant, based on using the sender's load as the denominator in the CMF computation. It is indeed giving slightly different results, but it's hard to tell whether on average they end up with better or worse load imbalances. At that stage I think it's an interesting option that deserves further consideration, so I've put it in the LBAF code as a `pmf_type` which can be either 0 (the fixed version using the max underload when it's over the average) or 1 (the "NS variant"). You can flip the switch in line 297 of `lbsRuntime.py`, right now this requires editing that line, but we should probably make this a command line option for more systematic investigation.

In any event I think we should also experiment with it on the `GossipLB` side.

ppebay: @lifflander @nlslatt can we resolve this thread and merge? Thanks

src/Model/lbsProcessor.py
```diff
@@ -251,28 +251,32 @@ def process_underload_message(self, msg):
 
     def compute_cmf_underloads(self, l_ave, pmf_type=0):
         """Compute CMF of underloads given an average load
+           tyoe 0: improved Gossip approac=h
```
nlslatt, line 254: Typos.

src/Model/lbsProcessor.py
```diff
@@ -251,28 +251,32 @@ def process_underload_message(self, msg):
...
-            factor = l_max if l_max > l_ave else l_ave
-                
-            # Initialize CMF value
-            sum_p = 0.
-            
-            # Iterate over all loads known to be underloaded
-            for l in loads:
-                # Update CMF
-                sum_p += 1. - factor * l
+            p_fac /= (l_max if l_max > l_ave else 1 / l_ave)
```
nlslatt, line 270: This looks suspicious to me, one option being `l_max` and the other being `1 / l_ave`.

src/Model/lbsProcessor.py
ppebay: Fixed!

nlslatt (changes requested):
One little change and it's fine, I think

src/Model/lbsProcessor.py
ppebay: Indeed !

nlslatt (approved):

PR #85, opened 2021-06-19
84 multiple correlation to determine lb predictors

ppebay:
Please note that for now, data input files are supposed to be co-located with the script and have names matching `in.*.dat`


lifflander (approved):
Looks great to me!

Issue #83, opened 2021-06-16
Terminate gracefully when  no edges are provided to a communication-based criterion


ppebay:
`LBAF` should not try to use criteria 2 or 3 when no communication edges were provided. It should error out cleanly with a proper message in this case, instead of crashing.

marcinwrobel1986:
Will be done with #80

Issue #82, opened 2021-06-16
Revamp the writer 


ppebay:
A few points to remember:
. no more "epochs" (e) but "phases" (p)
. the output file name should no longer contain the "time step" (nor the "phase id") 
. each record should start with the "phase id"

marcinwrobel1986:
Will be done with #80

Issue #81, opened 2021-06-16
Replace all instances of time_step with phase_id


ppebay:
The `time_step` terminology is outdated, as vt uses a more general concept of "phase".

The goal of this issue is to replace all instances of `time_step` with `phase_id`

marcinwrobel1986:
Will be done with #80

Issue #80, opened 2021-06-14
Change input/output format to JSON with brotli compression


lifflander:
The related issue in VT is: https://github.com/DARMA-tasking/vt/issues/1469 and the PR is: https://github.com/DARMA-tasking/vt/pull/1475

Here is the new JSON format we want to consume (one file per rank as before):
```json
{
    "phases": [
        {
            "tasks": [
                {
                    "time": 1.33514404296875e-05,
                    "subphases": [
                        {
                            "time": 1.33514404296875e-05,
                            "id": 0
                        }
                    ],
                    "resource": "cpu",
                    "object": 107374182400,
                    "node": 0
                },
                {
                    "time": 0.012618303298950195,
                    "subphases": [
                        {
                            "time": 0.012618303298950195,
                            "id": 0
                        }
                    ],
                    "resource": "cpu",
                    "object": 0,
                    "node": 0
                }
            ],
            "id": 0
        },
        {
            "tasks": [
                {
                    "time": 1.5974044799804688e-05,
                    "subphases": [
                        {
                            "time": 1.5974044799804688e-05,
                            "id": 0
                        }
                    ],
                    "resource": "cpu",
                    "object": 107374182400,
                    "node": 0
                },
                {
                    "time": 1.52587890625e-05,
                    "subphases": [
                        {
                            "time": 1.52587890625e-05,
                            "id": 0
                        }
                    ],
                    "resource": "cpu",
                    "object": 120259084288,
                    "node": 0
                },
                {
                    "time": 0.012134313583374023,
                    "subphases": [
                        {
                            "time": 0.012134313583374023,
                            "id": 0
                        }
                    ],
                    "resource": "cpu",
                    "object": 0,
                    "node": 0
                }
            ],
            "id": 2
        }
    ]
}
```

The JSON output might be compressed with Brotli or not (by default it will be). So we should use the python wrapped compression/decompression from this repo: https://github.com/google/brotli to perform the compression/decompression before reading/writing the LB stat file input/output.

marcinwrobel1986:
Ok, I am on it.
I think it would be better to implement before big architecture changes.

marcinwrobel1986:
@lifflander Do we want to have backward compatibility?

marcinwrobel1986:
@lifflander I have just pushed the converter we were talking yesterday about.
- uncompressed JSON is ca. 2.5 x bigger than CSV
- compressed JSON is ca. 6 x smalles than CSV

[...]

PR #86, opened 2021-06-21
Reading and writing a new format

marcinwrobel1986:
This PR resolves:
- #80 Change input/output format to JSON with brotli compression
- #81 Replace all instances of time_step with phase_id
- #82 Revamp the writer
- #83 Terminate gracefully when no edges are provided to a communication-based criterion

+ added converter from old CSV format to new JSON/compressed with Brotli


src/Applications/conf.yaml
```diff
@@ -14,7 +14,7 @@
 # threshold           [-r <rt>]   float     overload relative threshold
 # time_sampler_type   [-t <ts>]   str       object times sampler: <ts> in {uniform,lognormal}
 # weight_sampler_type [-w <ws>]   str       communications weights sampler: <cs> in {uniform,lognormal}
-# time_step           [-s <ts>]   int       time stepping for reading VT load logs
+# phase_id           [-s <ts>]   int       time stepping for reading VT load logs
```
nlslatt, line 17: This comment should be updated to reflect the change from timestep to phase.

src/IO/lbsLoadReaderVT.py
```diff
@@ -202,18 +139,56 @@ def read_iteration(self, n_p, time_step):
...
+                decompressed_dict = json.loads(compr_bytes.decode('utf-8'))
+
+        # defining phases from file
+        phases = decompressed_dict['phases']
+        # iterating over phases
+        for phase in phases:
+            phase_id = phase['id']
+            for task in phase['tasks']:
+                task_time = task.get('time')
+                task_object_id = task.get('entity').get('id')
```
nlslatt, line 176: Do we need to verify that the entity type is in fact `object`?

src/IO/lbsLoadReaderVT.py
```diff
@@ -202,18 +139,56 @@ def read_iteration(self, n_p, time_step):
...
+                    returned_dict.setdefault(phase_id, Processor(node_id))
+
+                    # Add object to processor
+                    returned_dict[phase_id].add_object(obj)
+
+                    # Print debug information when requested
+                    if self.verbose:
+                        print(f"{bcolors.HEADER}[LoadReaderVT] {bcolors.END}iteration = {phase_id}, "
+                              f"object id = {task_object_id}, time = {task_time}")
+
```
nlslatt, line 193: We're not reading in any communication edges.  Does ant part of LBAF require them?

src/IO/lbsLoadWriterVT.py
```diff
@@ -63,74 +65,84 @@ class LoadWriterVT:
...
+                    temp_dict[proc_id].append({'proc_id': proc_id, 'obj_id': obj_id, 'obj_time': obj_time})
+            except:
+                n_u += 1
+
+        dict_to_dump = dict()
+        dict_to_dump['phases'] = list()
+        for proc_id, others_list in temp_dict.items():
+            phase_dict = {'tasks': list(), 'id': proc_id}
+            for task in others_list:
+                task_dict = {'time': task['obj_time'], 'resource': 'cpu', 'object': task['obj_id']}
```
nlslatt, line 133: This does not appear to be consistent with the current schema (or the CSV to JSON converter).

utils/csv_2_json_data_converter.py
```diff
@@ -0,0 +1,171 @@
...
+        """ Method converts data to JSON and saves to output path
+        """
+        # Creating dictionary with right structure, which will be dumped
+        dict_to_dump = dict()
+        dict_to_dump['phases'] = list()
+        for proc_id, others_list in data_to_convert.items():
+            phase_dict = {'tasks': list(), 'id': proc_id}
+            for task in others_list:
+                task_dict = {'time': task['obj_time'], 'resource': 'cpu',
+                             'entity': {'id': task['obj_id'], 'type': 'object'}, 'node': node}
```
nlslatt, line 149: @lifflander Do we want to omit `home` or reverse engineer the entity ID (differently for CSVs that come from the release branch vs develop) so that the data is more complete?  With the `home` specified in all stats data imported into my LB testing/tuning driver in vt, I can delete a ton of code, but if we want to maintain compatibility with data that lacks `home` then I will need to retain all that extra functionality.

nlslatt (commented):
I made some comments, but was not able to provide a comprehensive review as I have limited experience with Python.  Will LBAF need to read or write communication data?  If so, there are a number of functions that will need to be extended.

src/Applications/conf.yaml
marcinwrobel1986: Thanks, for catching that up Nicole!

src/IO/lbsLoadReaderVT.py
marcinwrobel1986: There is a schema validator coming soon and we might need to have a quick chat regarding what else could it be.

src/IO/lbsLoadReaderVT.py
marcinwrobel1986: I haven't provide any additional functionality to LBAF, if it's lacking of some, that is another topic for the meeting.

src/IO/lbsLoadWriterVT.py
marcinwrobel1986: I will ask @lifflander Jonathan to send me the final data, then I would write a validator and make sure everytihng is up to date. Thanks!

PR #90, opened 2021-09-23
#89: add multi-regression scripts that predict

nlslatt:


marcinwrobel1986 (commented):
Hello @nlslatt ,
sorry for the delay, I can see here some C++ Python :)
It would be good to have any input data for that, so I can make sure it's still working + I would also restructure that a bit, so it would then be more frienndly to use.

nlslatt:
> Hello @nlslatt , sorry for the delay, I can see here some C++ Python :) It would be good to have any input data for that, so I can make sure it's still working + I would also restructure that a bit, so it would then be more frienndly to use.

Yeah, I have pretty much zero real Python experience so my C++ is bleeding through. I'll look into getting some data you can use to test it easily.

marcinwrobel1986:
> > Hello @nlslatt , sorry for the delay, I can see here some C++ Python :) It would be good to have any input data for that, so I can make sure it's still working + I would also restructure that a bit, so it would then be more frienndly to use.
> 
> Yeah, I have pretty much zero real Python experience so my C++ is bleeding through. I'll look into getting some data you can use to test it easily.

Nicole, it's probably 10^6 better than my C++ :)

scripts/multi_reg_and_predict_new_data.py
```diff
@@ -0,0 +1,163 @@
...
+import sys
+import re
+import numpy as np
+from sklearn import linear_model
+
+# This script trains on in.*.dat in the training directory passed on the
+# command-line and then predicts on in.*.dat in the working directory.
+
+def setupRegressors(files, bool_cols, y_col, rank_col, excluded, X, Y, ranks, regressor_list, n_obs):
+    first_row = True
```
marcinwrobel1986, line 12: Variable `first_row` is defined before for loop. This means that this variable will be True  just for first read `file_name` in `files`. Is it desired behaviour? Or maybe it should be after `for file_name in files:` but before `with open(file_name, ...`, then it would be True for each read file.

scripts/multi_reg_and_predict_new_data.py
```diff
@@ -0,0 +1,163 @@
...
+        with open(file_name, newline='') as csv_file:
+            # Open CSV reader with blank separators and numeric conversion
+            print("# Reading CSV file:", file_name)
+            csv_reader = csv.reader(csv_file,
+                                    delimiter=' ',
+                                    quoting=csv.QUOTE_NONNUMERIC)
+            # Iterate over file rows
+            for row in csv_reader:
+                # Compute record type from Boolean field and update as needed
+                r_type = sum([int(b) << i for i, b in enumerate(row) if i in bool_cols])
```
marcinwrobel1986, line 24: `<<` is Bitwise left shift, is that desired?

scripts/multi_reg_and_predict_new_data.py
```diff
@@ -0,0 +1,163 @@
...
+
+if len(sys.argv) != 2:
+    print("** ERROR: Specify training directory on the command-line")
+    sys.exit(1)
+
+training_dir = sys.argv[1];
+
+# Initialize global variables
+regressor_list = []
+n_obs = {}
```
marcinwrobel1986, line 82: `{}` could be set or dict. 
Empty set could be instantiated as `set()` and empty dict as `dict()`, that would be more explicit.
I would help with these two scripts. I just need some data, to be sure, to proceed correctly :)

marcinwrobel1986:
Hello again Nicole,
I have noticed you based your scripts on [multi_reg.py](https://github.com/DARMA-tasking/LB-analysis-framework/blob/develop/scripts/multi_reg.py) which was done by Philippe. That's a pitty I wasn't reviewing his PR. Let's talk about it on today's meeting.

marcinwrobel1986:
Hello @nlslatt Nicole and @ppebay Philippe,
if you woud have any input data, please let me know. Thanks!

ppebay:
This PR is superseded by PR #91

PR #97, opened 2021-10-31
#87: Ensure proper mirroring of edge storage

### LINKED ISSUE #87: Ensure proper mirroring of edge storage
By default we should assume that every edge `A -- x --> B` should be stored as `sent x to B` on `A ` **and** `received x from A` on  `B`. This is what pseudo-random samplers are doing; however VT traces do **not** store both entries. Therefore, we should by default ensure that such edges are properly "mirrored" in the LBAF data when read from such traces.

It could be useful to have the option (`OFF` by default) to deactivate this mirroring, to allow for, e.g., non-symmetrical load-balancing strategies; or, for the case where traces would already represent directed edges in both sides.

marcinwrobel1986:
- changed config file
- enchances populated from samplers option
- added communication reading

resolve #87

### DIFF
--- a/src/Applications/LBAF.py
+++ b/src/Applications/LBAF.py
@@ -308,15 +308,6 @@ class ggParameters:
             self.usage()
             sys.exit(1)
 
-        # # Checking if log dir exists, if not, checking if dir exists in project path
-        # if os.path.isdir(os.path.abspath(os.path.split(self.log_file)[0])):
-        #     self.log_file = os.path.abspath(self.log_file)
-        # elif os.path.isdir(os.path.join(project_path, os.path.split(self.log_file)[0])):
-        #     self.log_file = os.path.join(project_path, self.log_file)
-        # else:
-        #     print(f"{bcolors.ERR}LOG directory NOT FOUND!{bcolors.END}")
-        #     sys.exit(1)
-
         # Case when phases are populate from samplers not from log file
         if self.log_file is not None:
             # Checking if log dir exists, if not, checking if dir exists in project path
--- a/src/Applications/LBAF.py
+++ b/src/Applications/LBAF.py
@@ -308,14 +308,25 @@ class ggParameters:
             self.usage()
             sys.exit(1)
 
-        # Checking if log dir exists, if not, checking if dir exists in project path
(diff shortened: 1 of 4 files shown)


src/Applications/LBAF.py
```diff
@@ -308,14 +308,25 @@ def checks_after_init(self):
...
 
-        # Checking if log dir exists, if not, checking if dir exists in project path
-        if os.path.isdir(os.path.abspath(os.path.split(self.log_file)[0])):
-            self.log_file = os.path.abspath(self.log_file)
-        elif os.path.isdir(os.path.join(project_path, os.path.split(self.log_file)[0])):
-            self.log_file = os.path.join(project_path, self.log_file)
-        else:
-            print(f"{bcolors.ERR}LOG directory NOT FOUND!{bcolors.END}")
-            sys.exit(1)
+        # # Checking if log dir exists, if not, checking if dir exists in project path
```
ppebay, line 311: Why are these lines now commented out?

src/Applications/conf.yaml
```diff
@@ -12,9 +12,9 @@
 # n_rounds            [-k <nr>]   int       number of gossiping rounds
 # fanout              [-f <fo>]   int       gossiping fan-out value
 # threshold           [-r <rt>]   float     overload relative threshold
-# time_sampler_type   [-t <ts>]   str       object times sampler: <ts> in {uniform,lognormal}
-# weight_sampler_type [-w <ws>]   str       communications weights sampler: <cs> in {uniform,lognormal}
-# phase_id            [-s <ts>]   int       phase_id for reading VT load logs
+# time_sampler_type   [-t <ts>]   str       object times sampler: <ts> in {uniform,lognormal} E.g. lognormal,1.0,10.0
```
ppebay, line 15: please lowercase the `E` in `E.g.`

src/Applications/LBAF.py
marcinwrobel1986: done

src/Applications/conf.yaml
marcinwrobel1986: done

PR #100, opened 2021-11-04
#99: implemented non migratable objects and attendant method updates

### LINKED ISSUE #99: Create sentinel object attribute on each Rank
At Rank construction, one sentinel object should be created and assigned to `self.sentinel`

By default this object should have 0 load

Rank methods must be updated in order to take into account the load and communication related to that sentinel.
For example `get_load()` should also tally the load of the sentinel.

ppebay:
fixes #99


src/Model/lbsProcessor.py
```diff
@@ -82,14 +84,38 @@ def get_objects(self):
...
 
+        return self.objects.union([self.sentinel])
+
+    def get_migratable_objects(self):
+        """Return migratable objects assigned to processor
+        """
+
         return self.objects
 
+    def get_sentinel_object(self):
```
lifflander, line 95: In the new VT implementation, you will have a valid ID for every sentinel object. And you might have more than one, but you can collapse them if you want.

ppebay: Alright, so 2 follow-up questions:
* how is the valid ID generated. is there some logic that makes it predictable?
* shouldn't we then create a set of sentinel nodes (initially containing only one)?

src/Model/lbsProcessor.py
```diff
@@ -82,14 +84,38 @@ def get_objects(self):
...
+        """Return sentinel object assigned to processor
+        """
+
+        return self.sentinel
+
     def get_object_ids(self):
         """Return IDs of objects assigned to processor
         """
 
+        return [o.get_id() for o in self.objects + [self.sentinel]]
```
marcinwrobel1986, line 105: `Class 'set' does not define '__add__', so the '+' operator cannot be used on its instances `

I am not sure what's the intention here.
If you want to return a `list` with all objects ids including sentinel it should be:
```
obj_ids_list = [o.get_id() for o in self.objects]
obj_ids_list.append(self.sentinel.get_id())
return obj_ids_list
```
or
```
obj_ids_list = [o.get_id() for o in self.objects]
obj_ids_list.append(self.sentinel.index)
return obj_ids_list
```
If you want to return a `list` with objects ids and sentinel id as another `list` inside the first one it should be:
```
obj_ids_list = [o.get_id() for o in self.objects]
obj_ids_list.append([self.sentinel.get_id()])
return obj_ids_list
```
or
```
obj_ids_list = [o.get_id() for o in self.objects]
obj_ids_list.append([self.sentinel.index]`)
return obj_ids_list
```

src/Model/lbsProcessor.py
```diff
@@ -160,8 +186,20 @@ def get_load(self):
         """Return computed total load on processor
         """
 
+        return sum([o.get_time() for o in self.objects.union([self.sentinel])])
```
marcinwrobel1986, line 189: If the intention here is to return sum of time of all object including sentinel it should look like that:

```
proc_obj_with_sentinel = self.objects.add(self.sentinel)
return sum([o.get_time() for o in proc_obj_with_sentinel])
```

ppebay: @marcinwrobel1986 it should really be a set **union** so as to not replicate elements in the sum

src/Model/lbsProcessor.py
ppebay: Oh yes, thanks for pointing that, I erroneously treated the objects as `list` whereas it is a `set`, so as to not allow for element replication.

src/Model/lbsProcessor.py
marcinwrobel1986: So basically something like that:
```
self.objects.add(self.sentinel)
return sum([o.get_time() for o in self.objects])
```

marcinwrobel1986: Then maybe
```
return sum([o.get_time() for o in self.objects]) + self.sentinel.get_time()
```

ppebay: Sure by why is it better than what I did with the `union()`?  Is it more efficient in terms of alg. complexity?

marcinwrobel1986: I guess it's just readability, subjective thing, don't worry about that. You can keep it as you want Philippe :) I will understand that.

ppebay: I agree that the `union()` syntax is not very intuitive. It lends to believing that it is actually performing the union on the set that is given to it (while it is not). Not too sure which one reads easier, between both formulations

src/Model/lbsProcessor.py
```diff
@@ -41,26 +41,28 @@
...
         self.index = i
-        self.objects = set()
-        for obj in o:
-            self.add_object(obj)
+        self.migratable_objects = set()
+        for o in mo:
+            self.add_migratable_object(o)
+        self.sentinel_objects = set()
+        for o in so:
+            self.add_sentinelt(o)
```
lifflander, line 65: Spelling error?

lifflander (changes requested):

src/Model/lbsProcessor.py
ppebay: Yep thanks

lifflander (approved):

PR #116, opened 2021-12-06
#114: refactored parameters/criterion construction as needed

### LINKED ISSUE #114: Modify CriterionBase API to allow for passing of more parameters via config file
This is motivated in particular by. the fact that the total work criterion uses coefficients that should be modifiable via the LBAF config file at runtime

ppebay:
Fixes #114

### DIFF
--- a/src/Applications/conf.yaml
+++ b/src/Applications/conf.yaml
@@ -1,8 +1,9 @@
 # Docs
 # param_name_in_conf  [CL arg]    [type]    Description
 
-# criterion           [-c <tc>]   int(0-3)  transfer criterion: 0: lower total work, 1: tempered load only (default),
-                                                              # 2: strict localizer, 3: relaxed localizer
+# criterion           description of the object transfer criterion
+#    type : str in LowerTotalWork (default), TemperedLoad, StrictLocalizer, RelaxedLocalizer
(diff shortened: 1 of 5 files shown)


marcinwrobel1986:
Philippe @ppebay is that the final PR? So far I understand this would allow only one criterion, right?

src/Applications/conf.yaml
```diff
@@ -26,7 +26,11 @@
 # generate_multimedia [-g]        bool      generate multimedia
 # order_strategy                  str       arbitrary, element_id, fewest_migrations, small_objects, largest_objects
 
-criterion: 0
```
marcinwrobel1986, line 29: You should have changed the description/documentation in YAML file as well, so it corresponds to the different data format in conf file

criterion [-c <tc>]   int(0-3)  transfer criterion: 
- 0: lower total work, 
- 1: tempered load only (default),
- 2: strict localizer, 
- 3: relaxed localizer

ppebay: No. It is taking all existing criteria, but the description must be done as follows:
```
criterion:
  name: "LowerTotalWork"
  parameters:
    alpha: 0.
    beta: 0.001
```
They are to be called by name, instead of by index.

marcinwrobel1986 (approved):

PR #119, opened 2021-12-08
Removal of average load dependency

ppebay:
This needs to be thoroughly reviewed to ensure the modified algorithm is working properly


lifflander (approved):
Looks good!

Issue #126, opened 2021-12-19
Rename "weight" into "volume"


ppebay:
This is for communication between ranks in accord with more general practice in the literature

PR #128, opened 2021-12-22
#126: replaced weight with volume for communication nomenclature

ppebay:
Screen Shot 2021-12-22 at 11 03 32 AM

### DIFF
--- a/src/Applications/LBAF.py
+++ b/src/Applications/LBAF.py
@@ -89,8 +89,8 @@ class internalParameters:
         self.time_sampler_parameters = []
 
         # Object communication graph time sampler type and parameters
-        self.weight_sampler_type = None
-        self.weight_sampler_parameters = []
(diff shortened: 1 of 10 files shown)


src/Applications/LBAF.py
```diff
@@ -89,8 +89,8 @@ def __init__(self):
         self.time_sampler_parameters = []
 
         # Object communication graph time sampler type and parameters
-        self.weight_sampler_type = None
```
marcinwrobel1986, line 92: Please change that as well in the `conf.yml` file so the description is accurate as well

marcinwrobel1986: Sorry I see you changed it, I pulled the code and I don't have it locally :/

ppebay: Do you have it now?

marcinwrobel1986 (approved):

PR #131, opened 2021-12-23
#130 Modify load to become more generic work and make it abstract

### LINKED ISSUE #130: Modify "Load" to become "Work" and make it abstract
This is so that the concept of "Load" might be made to integrate non-loaded based metrics, in particular using communication.

ppebay:
Fixes #130

### DIFF
--- a/src/Applications/LBAF.py
+++ b/src/Applications/LBAF.py
@@ -72,6 +72,9 @@ class internalParameters:
     """
 
     def __init__(self):
+        # By default use load-only work model
+        self.work_model = {"name": "LoadOnly", "parameters": {}}
+
         # By default use tempered load criterion
         self.criterion = {"name": "TemperedLoad", "parameters": {}}
 
@@ -342,6 +345,7 @@ if __name__ == '__main__':
 
     # Instantiate runtime
     rt = Runtime(phase,
+                 params.work_model,
                  params.criterion,
                  params.order_strategy,
                  params.verbose)
--- a/src/Applications/LBAF.py
+++ b/src/Applications/LBAF.py
@@ -152,7 +152,11 @@ class internalParameters:
         """
 
         if os.path.splitext(conf_file)[-1] in [".yml", ".yaml"] and os.path.isfile(conf_file):
-            print(f"{bcolors.OKMSG}Config file {conf_file} FOUND!{bcolors.END}")
+            # Try to open configuration file
+            print(bcolors.HEADER
+                  + "[LBAF] "
+                  + bcolors.END
+                  + "Found configuration file {}".format(conf_file))
             try:
                 with open(conf_file, "rt") as config:
                     self.conf_file_found = True
--- a/src/Applications/LBAF.py
+++ b/src/Applications/LBAF.py
@@ -153,10 +153,7 @@ class internalParameters:
 
         if os.path.splitext(conf_file)[-1] in [".yml", ".yaml"] and os.path.isfile(conf_file):
             # Try to open configuration file
-            print(bcolors.HEADER
-                  + "[LBAF] "
-                  + bcolors.END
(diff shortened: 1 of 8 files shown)


src/Applications/LBAF.py
```diff
@@ -149,7 +152,11 @@ def get_conf_file(self, conf_file=os.path.join(project_path, "src", "Application
         """
 
         if os.path.splitext(conf_file)[-1] in [".yml", ".yaml"] and os.path.isfile(conf_file):
-            print(f"{bcolors.OKMSG}Config file {conf_file} FOUND!{bcolors.END}")
+            # Try to open configuration file
+            print(bcolors.HEADER
```
marcinwrobel1986, line 156: Maybe `fstring` here?
```Python
print(f"{bcolors.HEADER}[LBAF]{bcolors.END} Found configuration file {conf_file}")
```

ppebay: Yeah why not. But I don't like the `fstring` syntax. I find it much less clear.

ppebay: Done

src/Execution/lbsRuntime.py
```diff
@@ -76,13 +78,23 @@ def __init__(self, p, c: dict, order_strategy: str, v=False):
...
             self.phase = p
 
-        # Verbosity of runtime
-        self.verbose = v
+        # Instantiate work model
+        self.work_model = WorkModelBase.factory(
+            w.get("name"),
+            w.get("parameters", {}))
+        if not self.work_model:
+            print(bcolors.ERR
```
marcinwrobel1986, line 86: So, maybe we switch to fstring with every new commit :)?

src/Execution/lbsRuntime.py
```diff
@@ -93,21 +105,11 @@ def __init__(self, p, c: dict, order_strategy: str, v=False):
         _, l_min, self.average_load, l_max, l_var, _, _, l_imb = compute_function_statistics(
             self.phase.ranks,
             lambda x: x.get_load())
-        n_w, _, w_ave, w_max, w_var, _, _, w_imb = compute_function_statistics(
+        n_w, _, w_ave, w_max, _, _, _, _ = compute_function_statistics(
```
marcinwrobel1986, line 108: Just 3 out of 8 function outputs in use, maybe it's time to modify the output of `compute_function_statistics`

ppebay: Yes. We should return a structure and create a class for that (`lbsStatisticsContainer` or something like that). Another day. Please create another issue for that. Thanks.

src/Execution/lbsRuntime.py
ppebay: Better yet we should have a helper function called `print_error()`, another one called `print_warning()` and yet another one called `print_message()` that would generically deal with the properly-colored headings. In addition I would make the warnings have an orange-colored heading, the regular messages have a green heading (and keep the red for errors).

src/IO/lbsLoadReaderVT.py
```diff
@@ -193,7 +193,10 @@ def json_reader(self, returned_dict: dict, file_name: str, phase_ids, node_id: i
 
         # validate schema
         if SchemaValidator().is_valid(schema_to_validate=decompressed_dict):
-            print(f"{bcolors.OK}[LoadReaderVT] Valid JSON schema in {file_name}{bcolors.END}")
+            print(bcolors.HEADER
```
marcinwrobel1986, line 196:
```Python
print(f"{bcolors.HEADER}[LoadReaderVT]{bcolors.END} Valid JSON schema in {file_name}")
```

ppebay: Last one I am changing.
After that please check out #132

marcinwrobel1986 (approved):

PR #136, opened 2021-12-24
#134: implement concrete work based on combination of load and communication

### LINKED ISSUE #134: Implement concrete work model based on affine combination of load and communication
work = α + β * max(directed volume) + load

ppebay:
fixes #134

### DIFF
--- a/src/Applications/LBAF.py
+++ b/src/Applications/LBAF.py
@@ -218,6 +218,7 @@ class internalParameters:
         if isinstance(self.conf.get("order_strategy", None), str):
             self.order_strategy = self.conf.get("order_strategy", None)
 
+
 def parse_sampler(cmd_str):
     """Parse command line arguments specifying sampler type and input parameters
        Example: lognormal,1.0,10.0
--- a/src/Applications/LBAF.py
+++ b/src/Applications/LBAF.py
@@ -119,9 +119,6 @@ class internalParameters:
         # Base name to save computed object/rank mapping for VT
         self.map_file = None
 
-        # Decide whether actual destination loads should be computed
-        self.actual_dst_load = False
-
         # Decide whether Exodus output should be written
         self.exodus = False
 
@@ -371,7 +368,10 @@ if __name__ == '__main__':
 
     # Instantiate phase to VT file writer if started from a log file
     if params.log_file:
-        vt_writer = LoadWriterVT(phase, f"{output_stem}", output_dir=params.output_dir)
+        vt_writer = LoadWriterVT(
+            phase,
+            output_stem,
+            output_dir=params.output_dir)
         vt_writer.write()
 
     # If prefix parsed from command line
@@ -380,7 +380,7 @@ if __name__ == '__main__':
         ex_writer = WriterExodusII(
             phase,
             grid_map,
-            f"{output_stem}",
+            output_stem,
             output_dir=params.output_dir)
         ex_writer.write(
             rt.statistics,
--- a/src/Applications/LBAF.py
+++ b/src/Applications/LBAF.py
@@ -342,7 +342,7 @@ if __name__ == '__main__':
                               "initial rank loads",
                               params.verbose)
     print_function_statistics(phase.get_edges().values(),
-                              lambda x: x, "initial link volumes",
+                              lambda x: x, "initial sent volumes",
                               params.verbose)
 
     # Instantiate runtime
@@ -407,7 +407,10 @@ if __name__ == '__main__':
         lambda x: x.get_load(),
         "final rank loads",
         params.verbose)
-    print_function_statistics(phase.get_edges().values(), lambda x: x, "final link volumes", params.verbose)
+    print_function_statistics(
+        phase.get_edges().values(),
+        lambda x: x, "final sent volumes",
+        params.verbose)
 
     # Report on theoretically optimal statistics
     q, r = divmod(n_o, n_p)
--- a/src/Applications/LBAF.py
+++ b/src/Applications/LBAF.py
@@ -168,6 +168,7 @@ class internalParameters:
     def checks_after_init(self):
         """ Checks after initialization.
         """
+
         # Ensure that exactly one population strategy was chosen
         if (not (self.log_file or
                  (self.time_sampler_type and self.volume_sampler_type))
@@ -179,7 +180,7 @@ class internalParameters:
                   + bcolors.END)
             sys.exit(1)
 
-        # Case when phases are populate from samplers not from log file
+        # Case when phases are populated from samplers not from log file
         if self.log_file is not None:
             # Checking if log dir exists, if not, checking if dir exists in project path
             if os.path.isdir(os.path.abspath(os.path.split(self.log_file)[0])):
@@ -199,9 +200,12 @@ class internalParameters:
         """ Executed when config YAML file was found and checked
         """
 
+        # Assign parameters found in configuration file
         for param_key, param_val in self.conf.items():
-            if self.__dict__.get(param_key, "SomeRidiculousValue") != "SomeRidiculousValue":
+            if param_key in self.__dict__:
                 self.__dict__[param_key] = param_val
+
+        # Handle special values
         if isinstance(self.conf.get("x_procs", None), int) and self.conf.get("x_procs", 0) > 0:
             self.grid_size[0] = self.conf.get("x_procs", 0)
         if isinstance(self.conf.get("y_procs", None), int) and self.conf.get("y_procs", 0) > 0:
@@ -216,7 +220,6 @@ class internalParameters:
             self.communication_enabled = True
         if isinstance(self.conf.get("order_strategy", None), str):
             self.order_strategy = self.conf.get("order_strategy", None)
-
 
 def parse_sampler(cmd_str):
     """Parse command line arguments specifying sampler type and input parameters
--- a/src/Execution/lbsCriterionBase.py
+++ b/src/Execution/lbsCriterionBase.py
@@ -85,8 +85,6 @@ class CriterionBase:
         from src.Execution.lbsStrictLocalizingCriterion import StrictLocalizingCriterion
         from src.Execution.lbsRelaxedLocalizingCriterion import RelaxedLocalizingCriterion
 
-        criterion = locals()[criterion_name + "Criterion"]
-        return criterion(work_model, parameters)
         # Ensure that criterion name is valid
         try:
             # Instantiate and return object
--- a/src/Execution/lbsCriterionBase.py
+++ b/src/Execution/lbsCriterionBase.py
@@ -65,12 +65,15 @@ class CriterionBase:
                   + "*  ERROR: Could not create a criterion without a work model"
                   + bcolors.END)
             sys.exit(1)
+        self.get_work = lambda x: work_model.compute(x)
 
         # Criterion keeps internal references to ranks and edges
         print(bcolors.HEADER
               + "[CriterionBase] "
               + bcolors.END
-              + f"Created base criterion with work model: {work_model}")
+              + "Created base criterion with {} work model".format(
+                  str(type(work_model)).split('.')[-1][:-2]
+                  ))
 
     @staticmethod
     def factory(criterion_name, work_model, parameters={}):
@@ -82,6 +85,8 @@ class CriterionBase:
         from src.Execution.lbsStrictLocalizingCriterion import StrictLocalizingCriterion
         from src.Execution.lbsRelaxedLocalizingCriterion import RelaxedLocalizingCriterion
 
+        criterion = locals()[criterion_name + "Criterion"]
+        return criterion(work_model, parameters)
         # Ensure that criterion name is valid
         try:
             # Instantiate and return object
--- a/src/Execution/lbsCriterionBase.py
+++ b/src/Execution/lbsCriterionBase.py
@@ -53,7 +53,7 @@ class CriterionBase:
     """An abstract base class of optimization criteria for LBAF execution
     """
 
-    def __init__(self, work_model, parameters={}):
+    def __init__(self, work_model, parameters: dict=None):
         """Class constructor:
         work_model: a WorkModelBase instance
         parameters: optional parameters dictionary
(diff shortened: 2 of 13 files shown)


src/IO/lbsLoadReaderVT.py
```diff
@@ -132,53 +132,65 @@ def read(self, node_id: int, phase_id: int = -1, comm: bool = False) -> tuple:
...
+        """Read all the data in the range of ranks [0..n_p) for a given
         iteration `phase_id`. Collapse the iter_map dictionary from `read(..)`
         into a list of ranks to be returned for the given iteration.
         """
 
         # Create storage for ranks
-        procs = [None] * n_p
-        communication = dict()
+        rank_list = [None] * n_p
+        communications = {}
```
marcinwrobel1986, line 142: `{}` could define set as well. Please leave a `dict()`, that is explicit.

src/IO/lbsLoadReaderVT.py
```diff
@@ -191,67 +203,74 @@ def json_reader(self, returned_dict: dict, file_name: str, phase_ids, node_id: i
...
+                  + " Valid JSON schema in  {}".format(file_name))
         else:
             raise SyntaxError(f"{bcolors.ERR}[LoadReaderVT] Invalid JSON schema in {file_name}{bcolors.END}")
 
-        # defining phases from file
+        # Define phases from file
         phases = decompressed_dict["phases"]
-        comm_dict = dict()
-        # case for empty Rank
+        comm_dict = {}
```
marcinwrobel1986, line 217: This is implicit

src/IO/lbsLoadReaderVT.py
```diff
@@ -191,67 +203,74 @@ def json_reader(self, returned_dict: dict, file_name: str, phase_ids, node_id: i
...
-            # temporary communication list, to avoid duplicates in communication
-            temp_comm = list()
+            # Retrieve phase ID
             phase_id = phase["id"]
-            # adding communications to the object
-            communications = phase.get("communications", None)
-            # as communications is optional, there is a need to check if exists
-            if communications is not None and communications not in temp_comm:
+
+            # Ccreate communicator dictionary
```
marcinwrobel1986, line 228: Typo

src/IO/lbsLoadReaderVT.py
```diff
@@ -191,67 +203,74 @@ def json_reader(self, returned_dict: dict, file_name: str, phase_ids, node_id: i
...
+            comm_dict = {}
```
marcinwrobel1986, line 229: Implicit

src/IO/lbsLoadReaderVT.py
```diff
@@ -281,7 +299,7 @@ def csv_reader(self, returned_dict: dict, file_name: str, phase_id, node_id: int
         with open(file_name, 'r') as f:
             log = csv.reader(f, delimiter=',')
             # Iterate over rows of input file
-            comm_dict = dict()
+            comm_dict = {}
```
marcinwrobel1986, line 302: Implicit

src/Model/lbsRank.py
```diff
@@ -181,7 +181,7 @@ def get_load(self):
         """Return total load on rank
         """
 
-        return sum([o.get_time() for o in self.migratable_objects.union(self.sentinel_objects)])
+        return sum([o.get_time()for o in self.migratable_objects.union(self.sentinel_objects)])
```
marcinwrobel1986, line 184: space missing

src/Applications/LBAF.py
```diff
@@ -217,7 +221,6 @@ def parse_conf_file(self):
         if isinstance(self.conf.get("order_strategy", None), str):
             self.order_strategy = self.conf.get("order_strategy", None)
 
-
```
marcinwrobel1986, line 220: PEP 8 need 2 blank lines here, as this is function definition outside a class

src/IO/lbsLoadReaderVT.py
```diff
@@ -116,13 +116,13 @@ def read(self, node_id: int, phase_id: int = -1, comm: bool = False) -> tuple:
...
 
-        # Initialize storage
-        iter_map = dict()
-
-        # iter_map, comm = self.csv_reader(returned_dict=iter_map, file_name=file_name, phase_id=phase_id,
-        #                                  node_id=node_id)
-        iter_map, comm = self.json_reader(returned_dict=iter_map, file_name=file_name, phase_ids=phase_id,
-                                          node_id=node_id)
+        # Retrieve communications from JSON reader
+        iter_map = {}
```
marcinwrobel1986, line 120: Implicit

marcinwrobel1986:
Please make sure that work as intended, as I see many changes in the code.Thanks

src/IO/lbsLoadReaderVT.py
ppebay: Nope, when you use empty curly braces, it is not ambiguous, it is always an empty dict (and **not** an empty set). For an empty set you **must** use `set()`:
https://docs.python.org/2/tutorial/datastructures.html#sets

Furthermore, using `dict()` instead of `{}` results in much slower performance:
http://katrin-affolter.ch/Python/Python_dictionaries

src/IO/lbsLoadReaderVT.py
ppebay: Nope
```
>>> a={}
>>> type(a)
<class 'dict'>
>>> b={1,2,3}
>>> type(b)
<class 'set'>
```

src/IO/lbsLoadReaderVT.py
ppebay: Same comment as above

src/IO/lbsLoadReaderVT.py
ppebay: Same comment as above

ppebay:
Yes will do test, but it's still `WIP` so it's not finished yet. But yes I will add some tests.

src/IO/lbsLoadReaderVT.py
ppebay: Nope, it's explicit and more efficient perfomance-wise
```
>>> a={}
>>> type(a)
<class 'dict'>
>>> b={1,2,3}
>>> type(b)
<class 'set'>
```
http://katrin-affolter.ch/Python/Python_dictionaries

src/IO/lbsLoadReaderVT.py
marcinwrobel1986: ok, may fault, sorry, when it's empty, it will be a `dict`, but I would still insist regarding readability

ppebay:
@lifflander @PhilMiller 

In keeping with our earlier discussion, max directed volume is now used to compute work instead of bidirectional (summed) volume along edges.

Screen Shot 2021-12-26 at 7 22 16 PM
Screen Shot 2021-12-26 at 7 22 36 PM

ppebay:
@marcinwrobel1986 working as expected when `Work` is made to be equal to `Load` only (i.e. alpha and beta are both 0).

Screen Shot 2021-12-27 at 11 11 07 AM

src/Execution/lbsCriterionBase.py
```diff
@@ -46,78 +46,50 @@
...
-
+from src.Model.lbsWorkModelBase import WorkModelBase
 
 class CriterionBase:
     __metaclass__ = abc.ABCMeta
     """An abstract base class of optimization criteria for LBAF execution
     """
 
-    def __init__(self, ranks, edges, parameters=None):
+    def __init__(self, work_model, parameters={}):
```
marcinwrobel1986, line 56: danger, could behave not as intended
https://leimao.github.io/blog/Python-Default-Argument-Mutable-Object/

src/Execution/lbsCriterionBase.py
```diff
@@ -46,78 +46,50 @@
...
-              + "Assigned {} ranks and {} edges to base criterion".format(
-                  n_p,
-                  n_e))
+              + "Created base criterion with {} work model".format(
+                  str(type(work_model)).split('.')[-1][:-2]
+                  ))
 
     @staticmethod
-    def factory(criterion_name, ranks, edges, parameters=None):
+    def factory(criterion_name, work_model, parameters={}):
```
marcinwrobel1986, line 79: same as above
https://leimao.github.io/blog/Python-Default-Argument-Mutable-Object/

src/Execution/lbsCriterionBase.py
ppebay: Interesting. I did not know this _caveat_. Is there a PEP8 explicitly forbidding it? Otherwise I would leave it as it is, because it expresses the intent of what `parameters` should be, and forces subclass developers to follow that model.

src/Execution/lbsCriterionBase.py
ppebay: Cf. discussion above

src/Execution/lbsCriterionBase.py
marcinwrobel1986: PEP 8 is a style guide for Python Code
https://www.python.org/dev/peps/pep-0008/

Default parameter value is something different:
https://web.archive.org/web/20200221224620/http://effbot.org/zone/default-values.htm

You could use annotations in Python, they help to quickly read intentions as well works fine with IDE:
https://realpython.com/lessons/annotations/

```Python
def __init__(self, ranks, edges, parameters: dict = None):
```

ppebay: Okay. I read the article you sent above, but I still do not see clearly **what** problem it could cause exactly in our case. I would really like to understand what could go wrong here. Do you have an example in mind? Really curious here.

marcinwrobel1986: In this case maybe when called second time and more it could keep the `old` parameters and add something there instead initiate a new emty dictionary when called.

ppebay: For instance, would it be a good thing to do it in the superclass (the abstract one)?

Like, have `None` passed to superclass but initialize it to `{}` there if it's `None`?

marcinwrobel1986: No no, please take a look how to deal with that:
```Python
def myfunc(value=None):
    if value is None:
        value = []
```

ppebay: Yes I understand that but I find this ugly.
This is really a flaw of the language that it allows you to define a default value for a function parameter, and yet lets you alter that default inside the function.

marcinwrobel1986: Philippe it is what it is :), can't change that. Just trying to follow, to get what I want and avoid odd behavior.

ppebay: @marcinwrobel1986 I found a workaround that does not result in having to assign `{}` in case parameters are `None`:
```
        self.dst_work = get_dst_work_know_by_src if not (
            parameters and parameters.get(
                "actual_destination_work")) else get_actual_dst_work
```
and I have combined this with the annotations you suggested. I think the code looks alright like that. Please check it out.

src/Execution/lbsCriterionBase.py
```diff
@@ -46,78 +46,50 @@
...
-
+from src.Model.lbsWorkModelBase import WorkModelBase
 
 class CriterionBase:
     __metaclass__ = abc.ABCMeta
     """An abstract base class of optimization criteria for LBAF execution
     """
 
-    def __init__(self, ranks, edges, parameters=None):
+    def __init__(self, work_model, parameters: dict=None):
```
marcinwrobel1986, line 56: Ok, so now whay do we need `parameters` in this Class if it's not used?

src/Execution/lbsCriterionBase.py
```diff
@@ -46,78 +46,50 @@
...
-              + "Assigned {} ranks and {} edges to base criterion".format(
-                  n_p,
-                  n_e))
+              + "Created base criterion with {} work model".format(
+                  str(type(work_model)).split('.')[-1][:-2]
+                  ))
 
     @staticmethod
-    def factory(criterion_name, ranks, edges, parameters=None):
+    def factory(criterion_name, work_model, parameters={}):
```
marcinwrobel1986, line 79: same here, `dict` as default

src/Execution/lbsTemperedWorkCriterion.py
```diff
@@ -49,28 +49,36 @@
...
-        self.actual_dst_load = parameters.get("actual_destination_load", False)
+        
+        # Determine how destination work is to be computed
+        def get_dst_work_know_by_src(p_src, p_dst):
+            return p_src.get_known_work(p_dst)
+        def get_actual_dst_work(_, p_dst):
+            return self.get_work(p_dst)
+
+        # Retrieve releavant parameter when available
+        self.dst_work = get_dst_work_know_by_src if not (
```
marcinwrobel1986, line 73: Don't you have to pass arguments to the functions `get_dst_work_know_by_src` and `get_actual_dst_work`

marcinwrobel1986: Or you just want to `self.dst_work` to be a function?

src/Execution/lbsCriterionBase.py
ppebay: It is currently used by **some** of the concrete subclasses and in the future might become necessary for *all**

src/Execution/lbsTemperedWorkCriterion.py
ppebay: Yes, it has to be a function, used in the `compute()` method of the concrete criterion:
```
    def compute(self, obj: Object, p_src: Rank, p_dst: Rank) -> float:
        """Tempered work criterion based on L1 norm of works
        """

        # Criterion only uses object and rank works
        return self.get_work(p_src) - (
            self.dst_work(p_src, p_dst) + obj.get_time())
```
For efficiency's sake, we do not want this `if..else..` made at each invocation of `compute()` (as was the case before). Instead, we call the function that was properly initialized only once at construction time.

src/Execution/lbsCriterionBase.py
ppebay: Handled

src/Execution/lbsCriterionBase.py
ppebay: @marcinwrobel1986 did you see that change?

marcinwrobel1986 (approved):

Issue #122, opened 2021-12-09
Study and design a total work based CMF


ppebay:
This is to replace the load-based CMF computation below, to integrate communication weights:

<img width="675" alt="Screen Shot 2021-12-09 at 1 32 36 PM" src="https://user-images.githubusercontent.com/17439260/145397250-8dc1df5b-13f6-45c8-aa8f-f5edd6d06ae8.png">

ppebay:
Notes from 12/15/21 meeting

The objective function to be estimated must separate out "in" from "out" communication edges, and corresponding weights must be assigned to ranks and not to edges. This will allow, in particular, for a corresponding modification of the CMF as needed by this issue.

@PhilMiller can you please comment as you see fit?

ppebay:
@lifflander @PhilMiller 
Current change to the algorithm, based on assessing all **values** of the criterion of interest (rather than using a CMF) appears to be able to reduce communications to 0 when this is the objective function to be optimized:
<img width="810" alt="Screen Shot 2022-01-08 at 7 11 03 PM" src="https://user-images.githubusercontent.com/17439260/148655217-7c4426cd-cc29-4d00-95a1-2f0f1b985740.png">

ppebay:
@lifflander @PhilMiller 
The new method is based on the following reasoning:
```
                # Select best destination with respect to criterion
                c_max = 0.
                p_dst = None
                for p in targets.keys():
                    c = transfer_criterion.compute(o, p_src, p)
                    if c < 0.:
                        n_rejects += 1
                    if c >= c_max:
                        c_max = c
                        p_dst = p

                # Move to next object if no transfer was possible
                if not p_dst:
                    continue
```

PhilMiller:
That reads like it's deterministically choosing the best destination, which loses us the randomization that can let the algorithm move away from local minima.

ppebay:
@PhilMiller yes indeed, this method is there to be used as a reference point with deterministic transfer, to make sure whether our approach can attain a real optimum computed by the brute-force optimizer.

Following your comment I added an optional parameter that allows for the toggling between deterministic (optimization-based) and probabilistic (CMF-based) transfer strategies, in the `conf.yaml` file:
```
# deterministic_transfer    decide whether transfer must be deterministic
#    type: Boolean (default: False)
```

**N.B.:** for now, the probabilistic approach is still based on a load-only CMF, whereas the deterministic approach optimizes based on the work model

PR #138, opened 2021-12-27
#137 Generalize tempered criterion to be generically based on work

### LINKED ISSUE #137: Generalize "tempered" criterion to be generically based on "work"
Currently there still is a reference to the load of the object for transfer, but that should instead focus on its work. Therefore the notion of "work" must be extended to objects, as it is only defined for ranks (and it's obviously not naïvely additive, as it uses the max of two sums).

@PhilMiller @lifflander

ppebay:
fixes #137


ppebay:
@lifflander @PhilMiller
This criterion is now working as expected. For instance, when α=1, β=0, γ=0 (i.e. restoring the behavior of the old `TemperedLoad` criterion, we obtain as expected:

Screen Shot 2021-12-28 at 11 26 21 PM
(i.e. perfect load -- here, work -- balance)

In contrast when α=0, β=1, γ=0 (i.e when only taking communication volume into account), we obtain the following:
Screen Shot 2021-12-28 at 11 25 38 PM
Which is not entirely ideal yet, but this is because the CMF is not computed based on work yet (related to #122 )

marcinwrobel1986 (approved):

PR #139, opened 2021-12-29
#132: Create generic functions to print in standardized way

### LINKED ISSUE #132: Create generic functions to print messages, warnings, errors, in standardized way.
We should have a helper function called `print_error()`, another one called `print_warning()` and yet another one called `print_message()` that would generically deal with the properly-colored headings. In addition I would make the warnings have an orange-colored heading, the regular messages have a green heading (and keep the red for errors).

I would create a `lbsPrint.py` in `IO` with just the corresponding **static** functions.

marcinwrobel1986:
### This PR:
- adds logger instead of printing messages
- removes verbose option in config, instead there is `logging_level ` which could take `info`, `debug`, `warning` or `error` values
- logging level:
  - info - prints just info and errors
  - error - prints just errors
  - warning - prints warnings and errors
  - debug - prints everything (errors, info and debug)
- added colors for logging and new library `colorama`
- use `fstring` everywhere
- format code to PEP8 standard
- removed non used imports
- in `logger.ini` different formatters could be used as well to print e.g.:
  - just messages, 
  - loging level with module, function and messages
  - time with messages
  - other could be provide as well
- removed `bcolors` library from requirements
- provide logger for other Viewers
- moved 'utils' under the `src` - whole app is under `src`
- simplify testing

resolve #132


marcinwrobel1986:
@ppebay Please let me know what to do with `Viewers` in `Applications`, are they used? Shall we removed them? Shall I introduced logging there as well?
I will fix tomorrow CI as well (add utils to the path)

ppebay (approved):
Nice that you were able ro remove `bcolors`

PR #140, opened 2021-12-29
#122: add total work based CMF

### LINKED ISSUE #122: Study and design a total work based CMF

ppebay:
related to #122

### DIFF
--- a/src/Applications/conf.yaml
+++ b/src/Applications/conf.yaml
@@ -34,7 +34,7 @@ work_model:
   name: "AffineCombination"
   parameters:
     alpha: 1.
-    beta: 0.
+    beta: 1.
     gamma: 0.
 criterion:
   name: "Tempered"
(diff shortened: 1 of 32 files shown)


marcinwrobel1986 (approved):

src/Applications/conf.yaml
```diff
@@ -33,13 +35,14 @@
 work_model:
   name: "AffineCombination"
   parameters:
-    alpha: 1.
-    beta: 0.
+    alpha: 0.
```
lifflander, line 38: Do you want to commit the change to alpha/beta here?

lifflander (approved):
Overall, this looks great to me!

src/Applications/conf.yaml
ppebay: Thanks for catching this! 

I restored the default params to their previous values.

Issue #121, opened 2021-12-09
Study ordering strategies integrating communication weights


ppebay:
The goal of this issue is to study and discuss improved ordering strategies based on total work, to replace line 3 below:

<img width="393" alt="Screen Shot 2021-12-09 at 11 39 00 AM" src="https://user-images.githubusercontent.com/17439260/145381392-37fc139a-3c70-48da-a64f-ce00b55cba22.png">

ppebay:
For this more challenging mixed Load/Communication case:
```
alpha: 1.
beta: 1.
gamma: 0.
```
where the theoretical optimum is:
Number of generated arrangements: 262144
	minimax work: 4 for 24 arrangements
```
we only retrieve the following local optimum:

<img width="858" alt="Screen Shot 2022-02-13 at 2 12 56 PM" src="https://user-images.githubusercontent.com/17439260/153754864-14bef96e-d688-4b88-a049-84ff6454fdec.png">

ppebay:
With `decreasing_times` ordering (replacing `largest_objects`):
<img width="858" alt="Screen Shot 2022-02-13 at 6 27 09 PM" src="https://user-images.githubusercontent.com/17439260/153766904-56761a63-7ee8-4cf3-b2ee-5ad72172b87a.png">

ppebay:
With `increasing_times` ordering (new):
<img width="857" alt="Screen Shot 2022-02-13 at 6 28 45 PM" src="https://user-images.githubusercontent.com/17439260/153766962-7daad08f-beb9-4731-982a-b4deac8c39ff.png">

ppebay:
With `increasing_connectivity` ordering (new):
<img width="853" alt="Screen Shot 2022-02-13 at 9 10 46 PM" src="https://user-images.githubusercontent.com/17439260/153772954-3fa7b37e-e228-4b3f-a1c8-4442a0ffbd71.png">

PR #148, opened 2022-01-19
#122: implement work-based CMF method

ppebay:
Related to #122

### DIFF
--- a/src/Execution/lbsRuntime.py
+++ b/src/Execution/lbsRuntime.py
@@ -274,8 +274,8 @@ class Runtime:
 
                 else:
                     # Compute transfer CMF given information known to source
-                    p_cmf = p_src.compute_transfer_cmf(
-                        transfer_criterion, o, targets, True)
+                    p_cmf, c_values = p_src.compute_transfer_cmf(
+                        transfer_criterion, o, targets, False)
                     self.lgr.debug(self.ylw(f"\t  CMF = {p_cmf}"))
                     if not p_cmf:
                         continue
@@ -284,8 +284,7 @@ class Runtime:
                     p_dst = inverse_transform_sample(p_cmf)
 
                     # Decide whether proposed transfer passes criterion
-                    c_value = transfer_criterion.compute(o, p_src, p_dst) 
-                    if c_value < 0.:
+                    if c_values[p_dst] < 0.:
                         n_rejects += 1
                         continue
 
--- a/src/Execution/lbsRuntime.py
+++ b/src/Execution/lbsRuntime.py
@@ -254,7 +254,7 @@ class Runtime:
             while (o := next(obj_it, None)) is not None:
                 self.lgr.debug(self.ylw(f"\t* object {o.get_id()}:"))
                 # Initialize criterion value
-                c_value = 0.
+                c_value = -math.inf
                 
                 # Use deterministic or probabilistic transfer method
                 if deterministic_transfer:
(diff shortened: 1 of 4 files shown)


ppebay:
@lifflander @PhilMiller @nlslatt 

This is the best that can be achieved by the deterministic approach for a purely communication-driven criterion, due to the crystallization of two objects on rank 0:

Screen Shot 2022-01-19 at 9 23 30 PM

Note that this can be proven (by enumeration) that this is indeed the best solution that can be achieved **when allowing for only one object transfer at a time**

ppebay:
And now retrieving the same results (less fast -- 1 more iteration) with the new CMF based approach!
Screen Shot 2022-01-19 at 10 19 38 PM

src/Execution/lbsRuntime.py
```diff
@@ -264,25 +264,24 @@ def transfer_stage(self, transfer_criterion, deterministic_transfer):
...
+                    
                     # Move to next object if no transfer was possible
                     if not p_dst:
                         continue
 
                 else:
                     # Compute transfer CMF given information known to source
-                    p_cmf = p_src.compute_transfer_cmf()
+                    p_cmf = p_src.compute_transfer_cmf(
+                        transfer_criterion, o, targets, True)
```
nlslatt, line 278: Have you tried probabilistic with strict=True for pure loads? In TemperedLB, when I had tried omitting anything from the cmf that would later fail the criterion, it seemed to have trouble converging on a good solution. I'm wondering if you've seen the same thing in LBAF and if it might impact total work too.

ppebay: Yes same thing Nicole!
If I turn the strict flag to `True` it's not converging as well as when it's `False`.
I was a bit puzzled by that but am now reassured as you're describing a similar pattern with TemperedLB. Thanks!

nlslatt: Does that still happen with `actual_destination_load: True`?

ppebay: Yes!

nlslatt: Interesting. My initial thought had been that it was related to inadvertently overloading the most underloaded processors by not knowing what other processors are doing, but it sounds like you've ruled that out. Have you experimented with strict=True under different object orderings?

ppebay: Not yet because I wanted to explore that as part of #121 but it's likely these 2 issues (#122 and #121) are not entirely disjoint. 

What I find encouraging about what we are finding in the diagrams above, is that the criterion is working as expected, enforcing monotonicity (decreasing) of the objective function (total work). It's not converging to the absolute minimum due to the atomic transfers -- but at least it's not worsening the situation at any iteration.

nlslatt: Yes, that's definitely good. I'm very interested in where all this goes.

nlslatt (approved):
Looks good. One question about pre-existing code.

lifflander (approved):

src/Execution/lbsRuntime.py
```diff
@@ -264,29 +264,27 @@ def transfer_stage(self, transfer_criterion, deterministic_transfer):
                         c = transfer_criterion.compute(o, p_src, p)
                         if c < 0.:
                             n_rejects += 1
```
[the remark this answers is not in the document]
ppebay: This is wrong and I need to devise a better method :)

PR #158, opened 2022-02-04
#157: Make the rank-object-enumerator app use the conf.yaml file

### LINKED ISSUE #157: Make the rank-object-enumerator app use the conf.yaml file
Instead of taking as input a hard-coded block as is currently the case:
```
# Dictionary of objects
objects = (
    {"id": 0, "time": 1.0, "from": {}, "to": {5: 2.0}},
    {"id": 1, "time": 0.5, "from": {4: 2.0}, "to": {4: 1.0}},
    {"id": 2, "time": 0.5, "from": {3: 1.0}, "to": {}},
    {"id": 3, "time": 0.5, "from": {}, "to": {2: 1.0, 8: 0.5}},
    {"id": 4, "time": 0.5, "from": {1: 1.0}, "to": {1: 2.0}},
    {"id": 5, "time": 2.0, "from": {0: 2.0}, "to": {8: 2.0}},
    {"id": 6, "time": 1.0, "from": {7: 1.0, 8: 1.5}, "to": {}},
    {"id": 7, "time": 0.5, "from": {}, "to": {6: 1.0}},
    {"id": 8, "time": 1.5, "from": {3: 0.5, 5: 2.0}, "to": {6: 1.5}})

# Define number of ranks
n_ranks = 4

# Define work constants
alpha = 1.
beta = 1.
gamma = 0.
```
the `rank-object-enumerator.py` script (in `Applications`) should read `conf.yaml` and retrieve:
* `x_procs`, `y_procs`, and `z_procs`, from which it would compute `n_ranks` (the product of these 3 values);
* `alpha`, `beta`, `gamma` (and use them as such);
* `log_file`, `file_suffix`, and `phase_id` to retrieve the objects and their communications from VT trace data.

marcinwrobel1986:
- added reading from a config file
- assigned values from a config file instead of hardcoded values
- added function annotations
- fixed possible typo

close #157

### DIFF
--- a/src/Applications/rank-object-enumerator.py
+++ b/src/Applications/rank-object-enumerator.py
@@ -41,9 +41,21 @@
 #@HEADER
 #
 ###############################################################################
-import sys
-import math
 import itertools
+import math
+import os
+import sys
+
+import yaml
+
+
+def get_conf() -> dict:
+    """ Gets config from file and returns a dictionary. """
+    project_path = f"{os.sep}".join(os.path.abspath(__file__).split(os.sep)[:-3])
+    with open(os.path.join(project_path, "src", "Applications", "conf.yaml"), 'rt') as conf_file:
+        conf = yaml.safe_load(conf_file)
+    return conf
+
 
 # Dictionary of objects
 objects = (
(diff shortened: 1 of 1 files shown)


src/Applications/rank-object-enumerator.py
```diff
@@ -92,33 +110,33 @@ def compute_arrangement_works(arrangement, alpha, beta, gamma):
...
 
         # Compute communication volumes
-        works[rank] += beta * max(
+        works[rank] += beta_c * max(
             compute_volume(rank_objects, "from"),
             compute_volume(rank_objects, "to"))
 
         # Add constant
-        works[rank] + gamma
+        works[rank] += gamma_c
```
marcinwrobel1986, line 121: @ppebay could you please review this specific line. I changed it, because it doesn't have any impact before

ppebay: Yes this is good, thank you, that was a typo, but we never tested with nonzero `gamma`, so it would have failed. Thanks.

ppebay (approved):
Great work

Issue #161, opened 2022-02-05
Make the rank-object-enumerator app enumerate all optimal cases to a file


ppebay:
This is for comparison purposes with outputs of `LBAF` (and later of vt's `TemperedLB`).

ppebay:
Resolved by #163

PR #163, opened 2022-02-06
#160 #161 make recursiveness of transfer function configurable and make rank objecte numerator enumerate all optimal cases to a file

### LINKED ISSUE #160: Make recursiveness of transfer function configurable
This is following a suggestion by @lifflander in #152 so we can explore both ways easily (with or without recursion)

ppebay:
Resolve #160 #161


marcinwrobel1986 (approved):
@ppebay maybe we could save `optimal-arrangements.csv` to the `output` directory as we save all the results from LBAF. I could take it into consideration with my next PR

PR #172, opened 2022-02-14
#171: Have LBAF use a (new) output_file_stem variable for output files

### LINKED ISSUE #171: Have LBAF use a (new) output_file_stem variable for output files
Instead of constructing a file stem with the form `LBAF-n4-stats-i8-k4-f4-Tempered-True`, which is complicated and yet not descriptive enough, with a user-chosen stem, specified in the configuration file with a new variable called `output_file_stem` (with the appropriate suffixes `.vom`, `.ex2`, being added for the actual file names).

marcinwrobel1986:
### THIS PR:
- added `output_file_stem` key to the config file
- added support in LBAF for output file stem

close #171

### DIFF
--- a/src/Applications/LBAF.py
+++ b/src/Applications/LBAF.py
@@ -132,6 +132,9 @@ class internalParameters:
         # Output directory
         self.output_dir = None
 
+        # Output file steam
+        self.output_file_stem = None
+
         # Generate multimedia
         self.generate_multimedia = False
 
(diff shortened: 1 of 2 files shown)


src/Applications/LBAF.py
```diff
@@ -132,6 +132,9 @@ def __init__(self):
         # Output directory
         self.output_dir = None
 
+        # Output file steam
+        self.output_file_stem = None
```
ppebay, line 136: I am assuming this will result in a runtime error if the value is not provided, right?
Maybe we should have a default value for when the configuration file does not specify the output file stem.

ppebay (changes requested):

src/Applications/LBAF.py
marcinwrobel1986: If the `output_file_stem` in conf file is not provided, then it will use the old behaviour/naming convention

ppebay (approved):

PR #184, opened 2022-02-21
180 Make name of configuation file a mandatory parameter of LBAF

### LINKED ISSUE #180: Make name of configuation file a mandatory parameter of LBAF
Instead of the current mode of execution where LBAF is loading a configuration file called `conf.yaml`, there should be a mandatory parameter to be passed to it, e.g.,
```
python LBAF.py conf.yaml
```
in order to be able to maintain different live configurations, and alternate between them when running experiments.

This should really by **mandatory** (i.e. no default value).

marcinwrobel1986:
### THIS PR:

- added `config_file` as input parameter for `internalParameters` class

close #180 

@ppebay CI test is failing. I didn't make any changes which could make that happend. Config file is under `scripts/test_config.conf.yaml` please take a look. I also added print statement in test. It prints now found imbalance:
```shell
[LBAF] Process complete ###
@@@@@ FOUND IMBALANCE: 0.75 @@@@@
===> TEST FAILED!
```

### DIFF
--- a/scripts/test_lbaf.py
+++ b/scripts/test_lbaf.py
@@ -48,6 +48,7 @@ def run_tests():
     imbalance_file = f"/lbaf/output/imbalance.txt"
     with open(imbalance_file, 'r') as imb_file:
         imb_level = float(imb_file.read())
+        print(f"@@@@@ FOUND IMBALANCE: {imb_level} @@@@@")
         if imb_level < 0.000001:
             print(f"===> TEST PASSED!")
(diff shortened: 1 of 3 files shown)


ppebay:
@marcinwrobel1986 for CI it should have alpha = 1.0. and beta = gamma = 0.0
Is it the case?

ppebay (approved):
Changed value of beta (=0.0) for CI

PR #193, opened 2022-02-24
#183: Make computation of brute-force solution an option

### LINKED ISSUE #183: Make computation of brute-force solution an option
The configuration file should be allowed to specify a Boolean called `brute_force_optimization`, defaulting to `False` if absent,
so that the following section of `LBAF.py`:
```
    # Prepare input data for rank order enumerator
    objects = []
    for rank in phase.get_ranks():
        for o in rank.get_objects():
            entry = {
                "id": o.get_id(),
                "time": o.get_time(),
                "to": {},
                "from": {}}
            comm = o.get_communicator()
            if comm:
                for k, v in comm.get_sent().items():
                    entry["to"][k.get_id()] = v
                for k, v in comm.get_received().items():
                    entry["from"][k.get_id()] = v
            objects.append(entry)
    objects.sort(key=lambda x: x.get("id"))

    # Execute rank order enumerator and fetch optimal arrangements
    n_a, w_min_max, a_min_max = roe.compute_min_max_arrangements_work(
        objects)
    if n_a != n_ranks ** len(objects):
        lgr.error("Incorrect number of possible arrangements with repetition")
        sys.exit(1)
    lgr.info(f"Minimax work: {w_min_max:.4g} for {len(a_min_max)} optimal arrangements amongst {n_a}")

```
be executed only when said variable is `True`.

**N.B.:**
* in order for subsequent parts of the code to work where `a_min_max` is needed, it should be set to `[]` by default.
* @marcinwrobel1986 this can only be done after #182 is merged

marcinwrobel1986:
### This PR:
- added `brute_force_optimization` key to configuration
- added support for `brute_force_optimization` in LBAF (set False by default)
- set `a_min_max` to an empty list if `brute_force_optimization` is False

close #183


src/Applications/LBAF.py
```diff
@@ -379,31 +382,34 @@ def get_output_file_stem(params, n_ranks):
         "initial sent volumes",
         logger=lgr)
 
-    # Prepare input data for rank order enumerator
```
ppebay, line 382: Please add `info` level log comment here, saying that brute force optimization will **not** take place

marcinwrobel1986: @ppebay done

src/Applications/LBAF.py
```diff
@@ -379,31 +382,37 @@ def get_output_file_stem(params, n_ranks):
...
-
-    # Execute rank order enumerator and fetch optimal arrangements
-    n_a, w_min_max, a_min_max = roe.compute_min_max_arrangements_work(
-        objects)
-    if n_a != n_ranks ** len(objects):
-        lgr.error("Incorrect number of possible arrangements with repetition")
-        sys.exit(1)
-    lgr.info(f"Minimax work: {w_min_max:.4g} for {len(a_min_max)} optimal arrangements amongst {n_a}")
+    if params.brute_force_optimization:
+        lgr.info(f"Starting brute force optimization")
```
ppebay, line 386: No this is not what I meant: what I meant is that we want to indicate when we do **Not** do it.

When we do it there is no need to indicate it, because we do it from the brute-force results that are shown later in the log :D

marcinwrobel1986: I added both, should I remove the one when it's on?

src/Applications/LBAF.py
```diff
@@ -379,31 +382,37 @@ def get_output_file_stem(params, n_ranks):
...
+
+        # Execute rank order enumerator and fetch optimal arrangements
+        n_a, w_min_max, a_min_max = roe.compute_min_max_arrangements_work(
+            objects)
+        if n_a != n_ranks ** len(objects):
+            lgr.error("Incorrect number of possible arrangements with repetition")
+            sys.exit(1)
+        lgr.info(f"Minimax work: {w_min_max:.4g} for {len(a_min_max)} optimal arrangements amongst {n_a}")
+    else:
+        lgr.info(f"brute_force_optimization parameter in config file not set to True, "
```
marcinwrobel1986, line 413: @ppebay here is the info that it will not take place

ppebay: Please put instead "No brute force optimization performed". Thanks

ppebay (approved):

PR #207, opened 2022-03-24
#202: Investigate error in VT object map JSON format

### LINKED ISSUE #202: Investigate error in VT object map JSON format
There seems to be both missing (`home`) and unused (`type`) keys for objects. Please check and add rules to validator as needed.

@lifflander @nlslatt

marcinwrobel1986:
### THIS PR:
- updated synthetic stats
- added `home` in `entity` as obligatory
- changed home to the Rank where object exists during phase 0

Closes #202


marcinwrobel1986:
@nlslatt Could you please try to use the `synthetic_lb_stats` from this branch `202-investigate-error-in-VT-object-map-JSON-format` when you would find a minute, please? And if you have any problems, please report here. I will add whatever is missing. Thanks!

nlslatt:
I haven't run it through vt just yet, but the `home` should be the rank where the object exists during phase 0, so should not always be zero.

nlslatt:
@marcinwrobel1986 Please see my above feedback about the `home` if you missed it. I tried reading this data into vt and it didn't error out. However, it also didn't work correctly with my driver; I suspect this is at least in part due to the IDs not having the high bits that vt expects, but we won't know for sure until we make vt compatible with other numbering schemes for objects.

marcinwrobel1986:
Hello @nlslatt ,
thanks for the reminder. I will take a look at the latest data from Jonathan and make our synthetic data as similar as possible.

marcinwrobel1986:
@nlslatt Nicole, it should be good now. Please check when you have a moment. Thank you!

ppebay (commented):
Good for me but would prefer to have approval from @nlslatt as well before merging, thanks.

marcinwrobel1986:
> Good for me but would prefer to have approval from @nlslatt as well before merging, thanks.

Hi Philippe @ppebay ,
As spoke on yesterday's meeting. We can not do better than what is actually is, regarding data preparation. This is due to the fact, that we can not reproduce VT logic, specifically as this is a synthetic example, done on the paper by you and into json by me.
After I fixed the test I would say it is good to go.
We might think about creating an issue regarding data validator, I mean specifically what value apart from `object` can entity be.

nlslatt (approved):
This looks reasonable

Issue #229, opened 2022-04-22
Create strawman balancing algorithm for comparison purposes


ppebay:
This is to validate and compare existing actual algorithms

ppebay:
Resolved by #234

PR #234, opened 2022-04-26
#229 create strawman lb algorithm

ppebay:


src/lbaf/IO/lbsVTDataReader.py
```diff
@@ -52,7 +52,7 @@ def read(self, node_id: int, phase_id: int = -1, comm: bool = False) -> tuple:
         self.__logger.info(f"Reading {file_name} VT object map")
         if not os.path.isfile(file_name):
             self.__logger.error(f"File {file_name} does not exist.")
-            raise FileNotFoundError(f"File {file_name} not found!")
```
marcinwrobel1986, line 55: FileNotFoundError does:
- give a piece of information that raised error is `FIleNotFoundError`
- show user error message in this case: `f"File {file_name} not found!"`
- exits app

Plus it could be easily used with a unit test, when one wants to raise that error.

ppebay: We dot use exceptions in this code. Otherwise we would have to do it consistently everywhere.
We error out with a message.

marcinwrobel1986: I will be changing that successfully, so it can be tested by unit tests.

PR #239, opened 2022-05-02
WIP #236: Use pyzoltan to call Zoltan as a comparison algorithm in LBAF

### LINKED ISSUE #236: Use pyzoltan to call Zoltan as a comparison algorithm in LBAF
Let's add Zoltan as a strategy that can be called from LBAF.

marcinwrobel1986:
### THIS PR:
- added description in README.md regarding `PyZoltan` installation
- tested `PyZoltan` installation

Closes #236


marcinwrobel1986:
@ppebay Please take a look and once the installation is ok, you can change the assignee and continue with this PR

ppebay:
Thanks @marcinwrobel1986 but two questions:
- is the Zoltan binary actually linked?
- can you please add it to the dependencies and make sure it gets tested as part of CI? This will indeed now become an optional dependency.

marcinwrobel1986:
> Thanks @marcinwrobel1986 but two questions:
> 
> * is the Zoltan binary actually linked?
> * can you please add it to the dependencies and make sure it gets tested as part of CI? This will indeed now become an optional dependency.

Ok, good remark, we could test it in CI. For Ubuntu there is an easy way to install `Zoltan` with the package manager, but it would make sense if we do it in "universal" way, for both MacOS and Linux.

marcinwrobel1986:
Ad. 1. We are cloning PyZoltan, just to use the script provided there
Ad. 2. Please make sure you have `OpenMPI` installed
Here is the [documentation](https://pyzoltan.readthedocs.io/en/latest/start.html#building-and-linking-pyzoltan-on-osx-linux)
Some lines regardin MacOSX:
https://pyzoltan.readthedocs.io/en/latest/start.html#installing-mpi4py-and-zoltan-on-os-x

ppebay:
I have MPICH installed... so this pyzoltan can only work with OpenMPI?

marcinwrobel1986:
> I have MPICH installed... so this pyzoltan can only work with OpenMPI?

I am not an expert, but you can noticed that they want one to install OpenMPI
`sudo brew install open-mpi`

ppebay:
Yes I am installing it (via MacPorts actually) but I find it surprising that it would require a specific implementation of MPI (OpenMPI rather than MPICH).

ppebay:
Still not working on macOS, even with OpenMPI installed:
```
python setup.py install
--------------------------------------------------------------------------------
Unable to run mpic++ correctly, skipping parallel build
--------------------------------------------------------------------------------
  File "setup.py", line 423, in <module>
    setup_package()
  File "setup.py", line 369, in setup_package
    raise RuntimeError(
RuntimeError: There are no extension modules, Nothing to do!
```

marcinwrobel1986:
Philippe, maybe if you have it installed, then just:
```
# One needs to export ZOLTAN as an environment variable
export ZOLTAN=$INSTALL_PREFIX
```

`ZOLTAN` environment variable is very important, and it's used when installing `PyZoltan`

One needs to install `PyZoltan` requirements with (this is from PyZoltan repo):
```shell
pip install -r <some-directory>/pyzoltan/requirements.txt
```

Last step is to install `PyZoltan` itself:
```shell
pip install pyzoltan --no-build-isolation
```

ppebay:
Yes yes I did all of the above. Still failing.

lifflander:
> @lifflander if you have a chance, could you please try to check if pyzoltan's instructions are working with your version of macOS --- as you are probably using Homebrew rather than MacPorts?
> 
> Thanks!

I was able to get pyzoltan installed on the Mac. I think the key is having a proper version of OpenMPI installed. I ended up building everything with clang-5 and OpenMPI 3.

marcinwrobel1986:
Hello @ppebay I added readme instruction for Docker usage.

ppebay:
Please @tlamonthezie have a look at this

PR #242, opened 2022-05-04
#240: Implement and test validator-based parameter checking in LBAF

### LINKED ISSUE #240: Restore previous behavior of parameter checking in LBAF
We need to
- [ ] ensure that **no spurious** (useless) instance variables are created
- [ ] ensure that **only valid configuration keys** are processed when parsing the configuration file
- [ ] ensure that the keys give variable values with the **correct types**

So in addition to the allowed keys,  there should be the corresponding types.
`allowed_keys` would become a dict, where the keys are the current allowed keys, and the values are the corresponding types.

**In addition** in order to achieve the first constraint above, all these parameter values should no longer be stored as individual instance variables (created in `__init_()`or elsewhere), but rather there should be only a unique member variable to store all parameters, as  a dict, where the keys would be in the allowed keys and the types of the corresponding types. This would solve all constraints at once.

marcinwrobel1986:
### THIS PR:
- created configValidator
- created `from_data` and `from_samplers` for variants purposes
- followed recommendations from PR
- add some tests
- added configuration's allowed values
- change a few tests
- removed instance variables defined to None
- added configuration validation
- removed parsing sampler (no need anymore)
- improve parsing od configuration
- removed double checking from LBAF
- added some more checking in configurationValidator and improve error messages
- change test_config
- changed config
- made exodus optional
- added n_ranks as obligatory and to allowed config keys
- add checks against grid_size and n_rank
- removed unnecessary check
- moved code to under if when exodus is present
- modified all config tests
- changed acceptance test config

Closes #240


src/lbaf/Applications/conf.yaml
```diff
@@ -41,8 +41,9 @@
 #    parameters [list]         parameters e.g. 1.0,10.0 for lognormal
 
 # Specify input
-data_stem: "../data/synthetic_lb_data/data"
-phase_id: 0
+from_data:
```
ppebay, line 44: good idea

src/lbaf/Applications/conf.yaml
```diff
@@ -78,16 +79,17 @@ output_dir: ../../../output
...
-#  name: lognormal
-#  parameters:
-#    - 1.0
-#    - 10.0
-#volume_sampler:
-#  name: lognormal
-#  parameters:
-#    - 1.0
-#    - 10.0
+#from_samplers:
```
ppebay, line 82: good idea but make sure it's either/or: if both are turned ON or both are turned OFF ==> ERROR

src/lbaf/IO/configValidator.py
```diff
@@ -0,0 +1,123 @@
+from logging import Logger
+
+import yaml
+from schema import And, Optional, Or, Schema, Use
```
ppebay, line 4: please rather use `import schema as sc` and then use, e.g., `sc.Optional`, etc.

src/lbaf/IO/configValidator.py
```diff
@@ -0,0 +1,123 @@
...
+
+    @staticmethod
+    def is_valid(valid_schema: Schema, schema_to_validate: dict) -> bool:
+        """ Returns True is schema_to_validate is valid with self.valid_schema else False. """
+        is_valid = valid_schema.is_valid(schema_to_validate)
+        return is_valid
+
+    @staticmethod
+    def validate(valid_schema: Schema, schema_to_validate: dict):
+        """ Returns validated schema. """
```
ppebay, line 78: No `s` at the end of verb: use subjunctive form for descriptions (e.g., `Return validated schema.`)

src/lbaf/IO/configValidator.py
```diff
@@ -0,0 +1,123 @@
...
+        return valid_schema.validate(schema_to_validate)
+
+    def main(self):
+        """ Main routine for the config validation. """
+        # Validate Skeleton
```
ppebay, line 83: No capital `S` here.

src/lbaf/IO/configValidator.py
```diff
@@ -0,0 +1,123 @@
...
+        if self.is_valid(valid_schema=self.__skeleton, schema_to_validate=self.__config_to_validate):
+            self.__logger.info("Skeleton schema is VALID!")
```
ppebay, line 85: We never use all-cap words in LBAF messages. Change `VALID` to `valid`. Also, no exclamation mark.

src/lbaf/IO/configValidator.py
```diff
@@ -0,0 +1,123 @@
...
+        else:
+            self.__logger.error("Skeleton schema is INVALID!")
```
ppebay, line 87: Same comment as above

src/lbaf/IO/configValidator.py
```diff
@@ -0,0 +1,123 @@
...
+                self.__logger.info("from_data schema is VALID!")
+            else:
+                self.__logger.error("from_data schema is INVALID!")
+                self.validate(valid_schema=self.__from_data, schema_to_validate=from_data)
+        elif (from_samplers := self.__config_to_validate.get("from_samplers")) is not None:
+            self.__logger.info("Simulate from samplers was chosen.")
+            if self.is_valid(valid_schema=self.__from_samplers, schema_to_validate=from_samplers):
+                self.__logger.info("from_samplers schema is VALID!")
+            else:
+                self.__logger.error("from_samplers schema is INVALID!")
```
ppebay, line 103: Same. Remove all caps everywhere. No exclamation marks.

src/lbaf/IO/configValidator.py
```diff
@@ -0,0 +1,123 @@
...
+                                "deterministic_transfer": bool}}),
+            "BruteForce": Schema(
+                {"name": "BruteForce",
+                 Optional("parameters"): {"skip_transfer": bool}})
+        }
+        self.__logger = logger
+
+    @staticmethod
+    def is_valid(valid_schema: Schema, schema_to_validate: dict) -> bool:
+        """ Returns True is schema_to_validate is valid with self.valid_schema else False. """
```
ppebay, line 72: `is` ==> `if`

src/lbaf/IO/configValidator.py
```diff
@@ -0,0 +1,123 @@
...
+                                         error="uniform or lognormal needs to be chosen!"),
+                             "parameters": [float, float]},
+            "volume_sampler": {"name": And(str, Use(str.lower), lambda b: b in ("uniform", "lognormal"),
+                                           error="uniform or lognormal needs to be chosen!"),
+                               "parameters": [float, float]}
+        })
+        self.__algorithm = {
+            "InformAndTransfer": Schema(
+                {"name": "InformAndTransfer",
+                 "parameters": {"n_iterations": int, "n_rounds": int, "fanout": int, "order_strategy": And(
```
ppebay, line 54: Add extra newlines/indentations for legibility here

src/lbaf/IO/configValidator.py
```diff
@@ -0,0 +1,123 @@
...
+            Optional("output_dir"): str,
+            Optional("generate_multimedia"): bool
+        })
+        self.__from_data = Schema({"data_stem": str, "phase_id": int})
+        self.__from_samplers = Schema({
+            "n_objects": int,
+            "n_mapped_ranks": int,
+            "communication_degree": int,
+            "time_sampler": {"name": And(str, Use(str.lower), lambda a: a in ("uniform", "lognormal"),
+                                         error="uniform or lognormal needs to be chosen!"),
```
ppebay, line 45: no exclamation mark

src/lbaf/IO/configValidator.py
```diff
@@ -0,0 +1,123 @@
...
+                           "parameters": {"alpha": float,
+                                          "beta": float,
+                                          "gamma": float}},
+            "algorithm": {
+                "name": And(str, lambda d: d in ("InformAndTransfer", "BruteForce"),
+                            error="InformAndTransfer or BruteForce needs to be chosen!"),
+                Optional("parameters"): dict
+            },
+            "exodus": bool,
+            "x_procs": int,
```
ppebay, line 27: These variables should **not** be defined by default: `x_procs`, `y_procs`, `z_procs`. They are not used when `exodus` is OFF.

src/lbaf/IO/configValidator.py
```diff
@@ -0,0 +1,123 @@
...
+                self.validate(valid_schema=self.__algorithm.get(algorithm),
+                              schema_to_validate=self.__config_to_validate.get("algorithm"))
+
+
+if __name__ == "__main__":
+    with open("../Applications/conf.yaml", "rt") as config_file:
+        yaml_str = config_file.read()
+        configuration = yaml.safe_load(yaml_str)
+
+    ConfigValidator(config_to_validate=configuration, logger=logger()).main()
```
ppebay, line 123: Please use entire names : `configuration_validator` or `ConfigurationValidator`, but not `ConfigValidator`.

src/lbaf/IO/configValidator.py
```diff
@@ -0,0 +1,123 @@
...
+            self.__logger.info("Simulate from samplers was chosen.")
+            if self.is_valid(valid_schema=self.__from_samplers, schema_to_validate=from_samplers):
+                self.__logger.info("from_samplers schema is VALID!")
+            else:
+                self.__logger.error("from_samplers schema is INVALID!")
+                self.validate(valid_schema=self.__from_samplers, schema_to_validate=from_samplers)
+
+        # Validate Algorithm
+        if (algorithm := self.__config_to_validate.get("algorithm").get("name")) is not None:
+            self.__logger.info(f"Checking algorithm: {algorithm}")
```
ppebay, line 108: Message is not clear

src/lbaf/IO/configValidator.py
```diff
@@ -0,0 +1,123 @@
...
+                               "parameters": [float, float]}
+        })
+        self.__algorithm = {
+            "InformAndTransfer": Schema(
+                {"name": "InformAndTransfer",
+                 "parameters": {"n_iterations": int, "n_rounds": int, "fanout": int, "order_strategy": And(
+                     str, Use(str.lower), lambda e: e in ("arbitrary", "element_id", "increasing_times",
+                                                          "decreasing_times", "increasing_connectivity",
+                                                          "fewest_migrations", "small_objects"),
+                     error="arbitrary or element_id or increasing_times or decreasing_times or increasing_connectivity"
```
ppebay, line 58: This list of available strategies should not be hard-coded here. It should be defined somewhere else in a global list, akin for instance to the allowable keys.

src/lbaf/IO/configValidator.py
```diff
@@ -0,0 +1,123 @@
...
+        self.__from_data = Schema({"data_stem": str, "phase_id": int})
+        self.__from_samplers = Schema({
+            "n_objects": int,
+            "n_mapped_ranks": int,
+            "communication_degree": int,
+            "time_sampler": {"name": And(str, Use(str.lower), lambda a: a in ("uniform", "lognormal"),
+                                         error="uniform or lognormal needs to be chosen!"),
+                             "parameters": [float, float]},
+            "volume_sampler": {"name": And(str, Use(str.lower), lambda b: b in ("uniform", "lognormal"),
+                                           error="uniform or lognormal needs to be chosen!"),
```
ppebay, line 48: Same here: we should not hard-code the list of available options and at the same time the error message: this leads to database de-normalization as it becomes possible to have an error message not up-to-date with the list of available functions. This list must be defined only **once** and used by both the `lambda` and the `error`.

ppebay (changes requested):

src/lbaf/Applications/conf.yaml
marcinwrobel1986: it is defined already, I am working now on the tests, to check that 100%

src/lbaf/IO/configValidator.py
marcinwrobel1986: No need for that, now only used elements are imported, like `And, Optional, Or, Schema, Use`. If I would do `import schema as sc`, the whole module would be imported and there is no need for that.

ppebay: So yes indeed it's needed. We want to have the namespace.

src/lbaf/IO/configValidator.py
ppebay: Maybe as a first approach, just define the following tuple:
```
allowed_strategies = ("arbitrary", "element_id", "increasing_times","decreasing_times", "increasing_connectivity", "fewest_migrations", "small_objects")
```
and then you can test for `e in allowed_strategy` and for the message you can do `" or ".join(allowed_strategies)`

src/lbaf/IO/configValidator.py
marcinwrobel1986: yaml package will be removed, logging could also be removed, but `schema` will be the only package to use in this file now and later as well

ppebay: only there? Then ok

src/lbaf/IO/configValidator.py
marcinwrobel1986: Changed to:
```Python3
f"Checking algorithm schema of: {algorithm}"
```

src/lbaf/IO/configValidator.py
marcinwrobel1986: I will put the allowed values in `configurationValidator.py`, we could put them into other place if it exists.

src/lbaf/IO/configValidator.py
marcinwrobel1986: I will put the allowed values in `configurationValidator.py`, we could put them into other place if it exists.

tests/data/config/conf_correct_from_data_min_config.yml
```diff
@@ -0,0 +1,24 @@
...
+algorithm:
+  name: BruteForce
+
+# Specify output
+n_ranks: 4
+output_file_stem: output_file
+exodus:
+  x_procs: 2
+  y_procs: 2
+  z_procs: 1
```
ppebay, line 24: missing newline at end of file

src/lbaf/Applications/LBAF_app.py
```diff
@@ -43,56 +43,24 @@ def __init__(self, config_file: str):
         self.__allowed_config_keys = [
```
ppebay, line 43: Maybe make this a tuple instead of a list, for good form

src/lbaf/Applications/LBAF_app.py
```diff
@@ -160,29 +94,29 @@ def parse_conf_file(self):
...
+            if math.prod(self.grid_size) != self.n_ranks:
+                raise AssertionError(f"Grid size: {self.grid_size} from exodus is not equal to n_ranks: {self.n_ranks}")
+
+        # Parsing from data parameters if present
+        if self.configuration.get('from_data') is not None:
+            self.data_stem = self.configuration.get("from_data").get("data_stem")
+            self.phase_id = self.configuration.get("from_data").get("phase_id")
+
+        # Parsing sampling parameters if present
+        if self.configuration.get('from_samplers') is not None:
```
ppebay, line 111: double quotes

ppebay (changes requested):

tests/data/config/conf_correct_brute_force.yml
```diff
@@ -0,0 +1,29 @@
...
+  name: AffineCombination
+  parameters:
+    alpha: 1.
+    beta: 0.
+    gamma: 0.
+
+# Specify balancing algorithm
+brute_force_optimization: True
+algorithm:
+  name: BruteForce
```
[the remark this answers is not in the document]
marcinwrobel1986: It is included in testing algorithm schema:
```Python3
"BruteForce": Schema(
  {"name": "BruteForce",
   Optional("parameters"): {"skip_transfer": bool}})
```

src/lbaf/Applications/LBAF_app.py
```diff
@@ -160,29 +94,29 @@ def parse_conf_file(self):
...
-        if isinstance(self.configuration.get("volume_sampler_type", None), str):
-            self.volume_sampler_type, self.volume_sampler_parameters = self.parse_sampler(
-                self.configuration["volume_sampler_type"])
-
-        # Set logging level
-        ll = self.logging_level or "info"
+        # Parsing exodus if present
+        if self.configuration.get('exodus') is not None:
+            self.grid_size = []
+            for key in ["x_procs", "y_procs", "z_procs"]:
```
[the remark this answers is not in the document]
marcinwrobel1986: This is how it was before, changing to tuple

src/lbaf/Applications/LBAF_app.py
```diff
@@ -389,7 +273,11 @@ def main(self):
             vt_writer.write()
 
         # If prefix parsed from command line
-        if self.params.exodus:
+        if "exodus" in self.params.__dict__:
+            # Create mapping from rank to Cartesian grid
+            pgs = self.params.grid_size
```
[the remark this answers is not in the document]
marcinwrobel1986: I just put the code uder if, pgs is no need, removing

Issue #156, opened 2022-02-02
Create object-based ExodusII output


ppebay:
This should be done **in addition** to the existing rank-based output, **not** in its stead.

The goal would be to be able to show essentially the following:
![143924464-51667dfb-28ee-421d-ac58-5c14b1ed9dc0](https://user-images.githubusercontent.com/17439260/152226237-11883d25-ce5d-4f13-ba6a-335f6bd6c384.jpg)
**but** keeping our current approach of storing only the edge with maximum volume when there is one in each direction between two endoints.

distractor:
Not finished, but wrapped up.
@ppebay @marcinwrobel1986

ppebay:
@marcinwrobel1986 , @distractor won't have time to finish it this month.
Can you please pick up the issue and the branch from its current status?
Thanks

ppebay:
Resolved by #245

ppebay:
Resolved by PR #245

PR #245, opened 2022-05-07
#156 add object viewer

ppebay:
Object viewer is now complete.

It also works in sync with the rank viewer, hereby allowing for the visualization of objects inside their ranks:

Screen Shot 2022-05-07 at 2 39 48 PM
Screen Shot 2022-05-07 at 2 39 58 PM
Screen Shot 2022-05-07 at 2 40 06 PM
Screen Shot 2022-05-07 at 2 40 17 PM
Screen Shot 2022-05-07 at 2 40 31 PM

Some random jitter is added to the object coordinates, in order to reduce the edge overlap phenomenon when those are grid-aligned.


ppebay:
Pure-comm case (final optimum):
Screen Shot 2022-05-07 at 6 54 12 PM

We can see the nice "local cluster" on rank 0, with all inter-object edges amounting to 0.0 actual volume

marcinwrobel1986:
The changes below regarding formatting should be restore to previous state:
Zrzut ekranu z 2022-05-09 08-12-59
Max line lenght should be 120 chars.
Now, instead of 5 lines, there are 14 lines.

marcinwrobel1986:
Unit tests are failing, they need to be fixed and passed.

src/lbaf/Applications/LBAF_app.py
```diff
@@ -94,20 +94,22 @@ def parse_conf_file(self):
...
+        # Parse whether meshes must be generated
+        if (gm := self.configuration.get("generate_meshes")) is not None:
             self.grid_size = []
-            for key in ("x_procs", "y_procs", "z_procs"):
-                self.grid_size.append(self.configuration.get("exodus").get(key))
-            if math.prod(self.grid_size) != self.n_ranks:
-                raise AssertionError(f"Grid size: {self.grid_size} from exodus is not equal to n_ranks: {self.n_ranks}")
+            for key in ("x_ranks", "y_ranks", "z_ranks"):
+                self.grid_size.append(gm.get(key))
+            if math.prod(self.grid_size) < self.n_ranks:
```
marcinwrobel1986, line 102: shouldn't `math.prod(self.grid_size)` be equal to `self.n_ranks`? What if that would be bigger?

src/lbaf/Applications/LBAF_app.py
```diff
@@ -94,20 +94,22 @@ def parse_conf_file(self):
...
+                self.logger.error(f"Grid size: {self.grid_size} < {self.n_ranks}")
+                sys.exit(1)
```
marcinwrobel1986, line 104: As spoken, that is not a correct way to raise an error, it can not be tested. What was wrong with previous approach?

src/lbaf/Applications/rank_object_enumerator.py
```diff
@@ -266,7 +266,7 @@ def get_conf() -> dict:
     CONF = get_conf()
 
     # Define number of ranks
-    N_RANKS = CONF.get("x_procs") * CONF.get("y_procs") * CONF.get("z_procs")
+    N_RANKS = CONF.get("x_ranks") * CONF.get("y_ranks") * CONF.get("z_ranks")
```
marcinwrobel1986, line 269: This doesn't matter. As spoken rom should be removed and `x_ranks`, `y_ranks` and `z_ranks` are now under `generate_meshes` key in configuration, so it will not work anyway, because the returned vaule of `ranks` will be None.

src/lbaf/IO/lbsMeshWriter.py
```diff
@@ -0,0 +1,306 @@
...
+        self.__phase = p
+
+        # Retrieve number of mesh points and bail out early if empty set
+        if not (n_p := len(self.__phase.get_ranks())):
+            self.__logger.error("Empty list of ranks, cannot write a mesh file")
+            return
+        self.__n_p = n_p
+
+        # Ensure that an iterable of grid sizes was passed
+        if not isinstance(grid_size, (list, tuple)):
```
marcinwrobel1986, line 42: This is checked in `configurationValidator` no need for this check

src/lbaf/IO/lbsMeshWriter.py
```diff
@@ -0,0 +1,306 @@
...
+            self.__logger.error("Could not generate meshes by lack of a grid size iterable")
+            sys.exit(1)
+        self.__grid_size = grid_size
+        self.__logger.info(
+            f"Mapping {self.__n_p} ranks onto a {grid_size[0]}x"
+            f"{grid_size[1]}x{grid_size[2]} rectilinear grid")
+
+        # Ensure that a valid object jitter coefficient was passed
+        if not isinstance(object_jitter, float) or abs(object_jitter) >= 1.0:
```
marcinwrobel1986, line 51: This is checked in `configurationValidator` no need for this check

src/lbaf/IO/lbsMeshWriter.py
```diff
@@ -0,0 +1,306 @@
...
+            self.__logger.error(f"Could not generate meshes due to unsuitable object jitter coefficient: {object_jitter}")
+            sys.exit(1)
+        self.__object_jitter = object_jitter
+        
+        # Ensure that specified grid resolution is correct
+        if not isinstance(r, numbers.Number) or r <= 0.:
```
marcinwrobel1986, line 57: This is checked in `configurationValidator` no need for this check

src/lbaf/IO/lbsMeshWriter.py
```diff
@@ -0,0 +1,306 @@
...
+        self.__output_dir = output_dir
+        if self.__output_dir is not None:
+            self.__rank_file_name = os.path.join(
+                self.__output_dir,
+                self.__rank_file_name)
+            self.__object_file_name = os.path.join(
+                self.__output_dir,
+                self.__object_file_name)
+
+
```
marcinwrobel1986, line 74: Just one blank line. Two blank lines are expected between functions and classes and one blank line is expected between methods of a class.

src/lbaf/IO/lbsMeshWriter.py
```diff
@@ -0,0 +1,306 @@
...
+            return None, None, None
+
+        # Compute successive Euclidean divisions
+        k, r = divmod(flat_id, n01)
+        j, i = divmod(r, grid_sizes[0])
+
+        # Return Cartesian coordinates
+        return i, j, k
+
+
```
marcinwrobel1986, line 90: Just one blank line. Two blank lines are expected between functions and classes and one blank line is expected between methods of a class.

src/lbaf/IO/lbsMeshWriter.py
```diff
@@ -0,0 +1,306 @@
...
+import numbers
+import random
+import vtk
+
+from .lbsGridStreamer import GridStreamer
+from ..Model.lbsPhase import Phase
+
+
+class MeshWriter:
+    """A class to write LBAF results to mesh files via VTK layer
```
marcinwrobel1986, line 14: Doc strings quotations does not correspond the the one changes in other files.

src/lbaf/IO/lbsMeshWriter.py
```diff
@@ -0,0 +1,306 @@
...
+
+    def __init__(self, p: Phase, grid_size, object_jitter: float, f="lbs_out", r=1., output_dir=None, logger: Logger = None):
+        """ Class constructor:
+            p: Phase instance
+            grid_size: iterable containing grid sizes in each dimension
+            object_jitter: coefficient of random jitter with magnitude < 1
+            f: file name stem
+            r: grid_resolution value
+            output_dir: output directory
+        """
```
marcinwrobel1986, line 25: Doc strings quotations does not correspond the the one changes in other files.

src/lbaf/IO/lbsMeshWriter.py
```diff
@@ -0,0 +1,306 @@
...
+        k, r = divmod(flat_id, n01)
+        j, i = divmod(r, grid_sizes[0])
+
+        # Return Cartesian coordinates
+        return i, j, k
+
+
+    def write_rank_view_file(self, distributions: dict, statistics: dict):
+        """ Map ranks to grid and write ExodusII file
+        """
```
marcinwrobel1986, line 93: Doc strings quotations does not correspond the the one changes in other files. Or the method above.

src/lbaf/IO/lbsMeshWriter.py
```diff
@@ -0,0 +1,306 @@
...
+            self.__logger.info(f"Writing ExodusII file: {self.__rank_file_name}")
+            writer = vtk.vtkExodusIIWriter()
+            writer.SetFileName(self.__rank_file_name)
+            writer.SetInputConnection(streamer.Algorithm.GetOutputPort())
+            writer.WriteAllTimeStepsOn()
+            writer.Update()
+
+    def write_object_view_file(self, distributions: dict):
+        """ Map objects to grid and write ExodusII file
+        """
```
marcinwrobel1986, line 191: Doc strings quotations does not correspond the the one changes in other files.

src/lbaf/IO/lbsMeshWriter.py
```diff
@@ -0,0 +1,306 @@
...
+            file_name = f"{self.__object_file_name}_{iteration:02d}.vtp"
+            self.__logger.info(f"Writing VTP file: {file_name}")
+            writer = vtk.vtkXMLPolyDataWriter()
+            writer.SetFileName(file_name)
+            writer.SetInputData(pd_mesh)
+            writer.Update()
+
+    def write(self, distributions: dict, statistics: dict):
+        """ Write rank and object ExodusII files
+        """
```
marcinwrobel1986, line 301: Doc strings quotations does not correspond the the one changes in other files.

src/lbaf/Model/lbsPhase.py
```diff
@@ -113,9 +110,7 @@ def compute_edges(self):
         print_subset_statistics("Rank-local communication volume", v_local, "total volume", v_total, logger=self.__logger)
```
marcinwrobel1986, line 110: line too long

src/lbaf/Applications/conf.yaml
```diff
@@ -50,7 +50,7 @@ work_model:
   name: AffineCombination
   parameters:
     alpha: 1.
-    beta: 0.
```
marcinwrobel1986, line 53: I think we should not change values in configuration file, this spams the git and makes looking for changes more difficult.

src/lbaf/IO/configurationValidator.py
```diff
@@ -27,61 +33,99 @@ def __init__(self, config_to_validate: dict, logger: Logger):
         self.__config_to_validate = config_to_validate
         self.__skeleton = Schema({
             Or("from_data", "from_samplers", only_one=True): dict,
-            "work_model": {"name": And(str, lambda c: c in ALLOWED_WORK_MODELS,
-                                       error=f"{get_error_message(ALLOWED_WORK_MODELS)} needs to be chosen"),
-                           "parameters": {"alpha": float,
-                                          "beta": float,
-                                          "gamma": float}},
+            "work_model": {
```
marcinwrobel1986, line 36: formatting was readable, no need to add 60 extra lines to the file

src/lbaf/IO/configurationValidator.py
```diff
@@ -27,61 +33,99 @@ def __init__(self, config_to_validate: dict, logger: Logger):
...
+                lambda x: x > 0,
+                error="Should be of type 'int' and > 0"),
+            Optional("generate_meshes"): {
+                "x_ranks": And(int, lambda x: x > 0,
+                               error="Should be of type 'int' and > 0"),
+                "y_ranks": And(int, lambda x: x > 0,
+                               error="Should be of type 'int' and > 0"),
+                "z_ranks": And(int, lambda x: x > 0,
+                               error="Should be of type 'int' and > 0"),
+                "object_jitter": And(float, lambda x: abs(x) < 1.0,
```
marcinwrobel1986, line 63: If this is not optional it should be added to unit tests and acceptance test and it's configuration as well

src/lbaf/IO/lbsMeshWriter.py
```diff
@@ -0,0 +1,306 @@
...
+            self.__logger.error("Could not generate meshes by lack of a grid size iterable")
+            sys.exit(1)
+        self.__grid_size = grid_size
+        self.__logger.info(
+            f"Mapping {self.__n_p} ranks onto a {grid_size[0]}x"
+            f"{grid_size[1]}x{grid_size[2]} rectilinear grid")
+
+        # Ensure that a valid object jitter coefficient was passed
+        if not isinstance(object_jitter, float) or abs(object_jitter) >= 1.0:
+            self.__logger.error(f"Could not generate meshes due to unsuitable object jitter coefficient: {object_jitter}")
```
marcinwrobel1986, line 52: line longer than 120 chars

src/lbaf/IO/lbsMeshWriter.py
```diff
@@ -0,0 +1,306 @@
...
+
+        # Retrieve number of mesh points and bail out early if empty set
+        n_o = self.__phase.get_number_of_objects()
+        if not n_o:
+            self.__logger.error("Empty list of objects, cannot write a mesh file")
+            return
+
+        # Number of edges is fixed due to vtkExodusIIWriter limitation
+        n_e = int(n_o * (n_o - 1) / 2)
+        self.__logger.info(f"Creating object view mesh with {n_o} points, {n_e} edges, and jitter factor: {self.__object_jitter}")
```
marcinwrobel1986, line 201: line longer than 120 chars

src/lbaf/Utils/csv_2_json_data_converter.py
```diff
@@ -96,74 +93,89 @@ def _get_files_for_conversion(self) -> list:
...
-        """
+        """ Get lists of files to convert. Iterate over it and converts each file."""
         files_to_convert = self._get_files_for_conversion()
         for file in files_to_convert:
             self._convert_file(file_path=file)
 
 
-if __name__ == '__main__':
-    Csv2JsonConverter(dir_path='data/vt_example_lb_data', compressed=False).main()
+if __name__ == "__main__":
```
marcinwrobel1986, line 180: Does VT produces any CSV data anymore? I think `csv_2_json_data_converter.py` was created with purpose to used against `old` well known csv data in order to cenvert them to a new format. At the moment we use `hand made data` for testing purposes and it seems not to be relevant anymore.

tests/data/config/conf_correct_001.yml
```diff
@@ -29,7 +29,7 @@ terminal_background: light
...
 output_dir: ../../../output
 output_file_stem: output_file
-exodus:
-  x_procs: 2
-  y_procs: 2
-  z_procs: 1
+generate_meshes:
+  x_ranks: 2
+  y_ranks: 2
+  z_ranks: 1
```
marcinwrobel1986, line 35: missing `object_jitter` if not optional

tests/data/config/conf_correct_002.yml
```diff
@@ -24,10 +24,10 @@ terminal_background: light
...
 output_dir: ../../../output
 output_file_stem: output_file
-exodus:
-  x_procs: 2
-  y_procs: 2
-  z_procs: 1
+generate_meshes:
+  x_ranks: 2
+  y_ranks: 2
+  z_ranks: 1
```
marcinwrobel1986, line 30: missing `object_jitter` if not optional

tests/data/config/conf_correct_003.yml
```diff
@@ -31,7 +31,7 @@ terminal_background: light
...
 output_dir: ../../../output
 output_file_stem: output_file
-exodus:
-  x_procs: 2
-  y_procs: 2
-  z_procs: 1
+generate_meshes:
+  x_ranks: 2
+  y_ranks: 2
+  z_ranks: 1
```
marcinwrobel1986, line 37: missing `object_jitter` if not optional

tests/data/config/conf_correct_brute_force.yml
```diff
@@ -23,7 +23,7 @@ terminal_background: light
...
 output_dir: ../../../output
 output_file_stem: output_file
-exodus:
-  x_procs: 2
-  y_procs: 2
-  z_procs: 1
+generate_meshes:
+  x_ranks: 2
+  y_ranks: 2
+  z_ranks: 1
```
marcinwrobel1986, line 29: missing `object_jitter` if not optional

tests/data/config/conf_correct_from_samplers_no_ll.yml
```diff
@@ -24,10 +24,10 @@ terminal_background: light
...
 output_dir: ../../../output
 output_file_stem: output_file
-exodus:
-  x_procs: 2
-  y_procs: 2
-  z_procs: 1
+generate_meshes:
+  x_ranks: 2
+  y_ranks: 2
+  z_ranks: 1
```
marcinwrobel1986, line 30: missing `object_jitter` if not optional

tests/data/config/conf_wrong_data_and_sampling.yml
```diff
@@ -29,10 +29,10 @@ terminal_background: light
...
 output_dir: ../../../output
 output_file_stem: output_file
-exodus:
-  x_procs: 2
-  y_procs: 2
-  z_procs: 1
+generate_meshes:
+  x_ranks: 2
+  y_ranks: 2
+  z_ranks: 1
```
marcinwrobel1986, line 35: missing `object_jitter` if not optional

tests/data/config/conf_wrong_from_data_phase_name.yml
```diff
@@ -29,7 +29,7 @@ terminal_background: light
...
 output_dir: ../../../output
 output_file_stem: output_file
-exodus:
-  x_procs: 2
-  y_procs: 2
-  z_procs: 1
+generate_meshes:
+  x_ranks: 2
+  y_ranks: 2
+  z_ranks: 1
```
marcinwrobel1986, line 35: missing `object_jitter` if not optional

tests/data/config/conf_wrong_from_data_phase_type.yml
```diff
@@ -29,7 +29,7 @@ terminal_background: light
...
 output_file_stem: output_file
-exodus:
-  x_procs: 2
-  y_procs: 2
-  z_procs: 1
\ No newline at end of file
+generate_meshes:
+  x_ranks: 2
+  y_ranks: 2
+  z_ranks: 1
```
marcinwrobel1986, line 35: missing `object_jitter` if not optional

tests/data/config/conf_wrong_from_samplers_time_sampler_001.yml
```diff
@@ -24,10 +24,10 @@ terminal_background: light
...
 output_dir: ../../../output
 output_file_stem: output_file
-exodus:
-  x_procs: 2
-  y_procs: 2
-  z_procs: 1
+generate_meshes:
+  x_ranks: 2
+  y_ranks: 2
+  z_ranks: 1
```
marcinwrobel1986, line 30: missing `object_jitter` if not optional

tests/data/config/conf_wrong_from_samplers_time_sampler_002.yml
```diff
@@ -24,10 +24,10 @@ terminal_background: light
...
 output_dir: ../../../output
 output_file_stem: output_file
-exodus:
-  x_procs: 2
-  y_procs: 2
-  z_procs: 1
+generate_meshes:
+  x_ranks: 2
+  y_ranks: 2
+  z_ranks: 1
```
marcinwrobel1986, line 30: missing `object_jitter` if not optional

tests/data/config/conf_wrong_from_samplers_time_sampler_003.yml
```diff
@@ -24,10 +24,10 @@ terminal_background: light
...
 output_dir: ../../../output
 output_file_stem: output_file
-exodus:
-  x_procs: 2
-  y_procs: 2
-  z_procs: 1
+generate_meshes:
+  x_ranks: 2
+  y_ranks: 2
+  z_ranks: 1
```
marcinwrobel1986, line 30: missing `object_jitter` if not optional

tests/data/config/conf_wrong_from_samplers_time_sampler_004.yml
```diff
@@ -24,10 +24,10 @@ terminal_background: light
...
 output_dir: ../../../output
 output_file_stem: output_file
-exodus:
-  x_procs: 2
-  y_procs: 2
-  z_procs: 1
+generate_meshes:
+  x_ranks: 2
+  y_ranks: 2
+  z_ranks: 1
```
marcinwrobel1986, line 30: missing `object_jitter` if not optional

tests/data/config/conf_wrong_from_samplers_time_sampler_005.yml
```diff
@@ -24,10 +24,10 @@ terminal_background: light
...
 output_dir: ../../../output
 output_file_stem: output_file
-exodus:
-  x_procs: 2
-  y_procs: 2
-  z_procs: 1
+generate_meshes:
+  x_ranks: 2
+  y_ranks: 2
+  z_ranks: 1
```
marcinwrobel1986, line 30: missing `object_jitter` if not optional

tests/data/config/conf_wrong_missing_from_data_param.yml
```diff
@@ -28,7 +28,7 @@ terminal_background: light
...
 output_dir: ../../../output
 output_file_stem: output_file
-exodus:
-  x_procs: 2
-  y_procs: 2
-  z_procs: 1
+generate_meshes:
+  x_ranks: 2
+  y_ranks: 2
+  z_ranks: 1
```
marcinwrobel1986, line 34: missing `object_jitter` if not optional

tests/data/config/conf_wrong_no_data_and_sampling.yml
```diff
@@ -24,7 +24,7 @@ terminal_background: light
...
 output_dir: ../../../output
 output_file_stem: output_file
-exodus:
-  x_procs: 2
-  y_procs: 2
-  z_procs: 1
+generate_meshes:
+  x_ranks: 2
+  y_ranks: 2
+  z_ranks: 1
```
marcinwrobel1986, line 30: missing `object_jitter` if not optional

tests/data/config/conf_wrong_work_model_missing.yml
```diff
@@ -21,7 +21,7 @@ terminal_background: light
...
 output_dir: ../../../output
 output_file_stem: output_file
-exodus:
-  x_procs: 2
-  y_procs: 2
-  z_procs: 1
+generate_meshes:
+  x_ranks: 2
+  y_ranks: 2
+  z_ranks: 1
```
marcinwrobel1986, line 27: missing `object_jitter` if not optional

tests/data/config/conf_wrong_work_model_name.yml
```diff
@@ -29,7 +29,7 @@ terminal_background: light
...
 output_dir: ../../../output
 output_file_stem: output_file
-exodus:
-  x_procs: 2
-  y_procs: 2
-  z_procs: 1
+generate_meshes:
+  x_ranks: 2
+  y_ranks: 2
+  z_ranks: 1
```
marcinwrobel1986, line 35: missing `object_jitter` if not optional

tests/data/config/conf_wrong_work_model_parameters_missing.yml
```diff
@@ -28,7 +28,7 @@ terminal_background: light
...
 output_dir: ../../../output
 output_file_stem: output_file
-exodus:
-  x_procs: 2
-  y_procs: 2
-  z_procs: 1
+generate_meshes:
+  x_ranks: 2
+  y_ranks: 2
+  z_ranks: 1
```
marcinwrobel1986, line 34: missing `object_jitter` if not optional

tests/data/config/conf_wrong_work_model_parameters_type.yml
```diff
@@ -29,7 +29,7 @@ terminal_background: light
...
 output_dir: ../../../output
 output_file_stem: output_file
-exodus:
-  x_procs: 2
-  y_procs: 2
-  z_procs: 1
+generate_meshes:
+  x_ranks: 2
+  y_ranks: 2
+  z_ranks: 1
```
marcinwrobel1986, line 35: missing `object_jitter` if not optional

marcinwrobel1986 (changes requested):
Unit tests must pass.
New changes into configuration should be then passed to acceptance tests configuration and checked with CI.

src/lbaf/Utils/csv_2_json_data_converter.py
nlslatt: It doesn't produce CSV data anymore, but we do have some archived so this converter could potentially still be used.

src/lbaf/Applications/LBAF_app.py
ppebay: It _can_ be bigger. We need to have **at least** enough grid vertices to place all ranks. For example say we have 7 ranks, we can place them on a 2x2x2 grid (or 4x2x1, or even 8x1x1). But we cannot place them on a 3x2x1 grid.

src/lbaf/Applications/rank_object_enumerator.py
ppebay: `procs` is older terminology. We now use `ranks` instead everywhere. Those `_procs` were leftover from initial implementations where the notion of "processor" was used.

src/lbaf/IO/configurationValidator.py
```diff
@@ -102,35 +146,49 @@ def validate(valid_schema: Schema, schema_to_validate: dict):
     def main(self):
         """ Main routine for the config validation. """
         # Validate skeleton
-        if self.is_valid(valid_schema=self.__skeleton, schema_to_validate=self.__config_to_validate):
+        if self.is_valid(
+            valid_schema=self.__skeleton,
```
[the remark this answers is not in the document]
ppebay: Does it give an error?

src/lbaf/IO/configurationValidator.py
```diff
@@ -102,35 +146,49 @@ def validate(valid_schema: Schema, schema_to_validate: dict):
...
+            self.validate(
+                valid_schema=self.__skeleton,
+                schema_to_validate=self.__config_to_validate)
 
         # Validate from_data/from_samplers
         if (from_data := self.__config_to_validate.get("from_data")) is not None:
             self.__logger.info("Reading from data was chosen.")
-            if self.is_valid(valid_schema=self.__from_data, schema_to_validate=from_data):
+            if self.is_valid(
+                valid_schema=self.__from_data,
```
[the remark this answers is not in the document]
ppebay: Does it give an error?

src/lbaf/IO/configurationValidator.py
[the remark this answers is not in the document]
marcinwrobel1986: Python interpreter can parse it, but this is an error: `PEP 8: E125 continuation line with same indent as next logical line`,
it looks like they are all arguments of the same function.
```Python3
        if self.is_valid(
            valid_schema=self.__skeleton,
            schema_to_validate=self.__config_to_validate):
            self.__logger.info("Skeleton schema is valid")
```
We have line length up to 120 chars. It could be easily visible even with resolution 1280x1024 or 1280x720.
Should be as it was:
```Python3
        if self.is_valid(valid_schema=self.__skeleton, schema_to_validate=self.__config_to_validate):
            self.__logger.info("Skeleton schema is valid")
```

src/lbaf/IO/configurationValidator.py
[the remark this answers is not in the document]
marcinwrobel1986: Python interpreter can parse it, but this is an error: `PEP 8: E125 continuation line with same indent as next logical line`,
it looks like they are all arguments of the same function.
```Python3
            if self.is_valid(
                valid_schema=self.__from_data,
                schema_to_validate=from_data):
                self.__logger.info("from_data schema is valid")
```
We have line length up to 120 chars. It could be easily visible even with resolution 1280x1024 or 1280x720.
Should be as it was:
```Python3
            if self.is_valid(valid_schema=self.__from_data, schema_to_validate=from_data):
                self.__logger.info("from_data schema is valid")
```

src/lbaf/Applications/rank_object_enumerator.py
marcinwrobel1986: I am not saying, it's not right, just as we agreed, that roe will be removed, so I see no point of editing roe anymore. If I remember well, one or more functions should stay and the rest should be removed, right?

src/lbaf/Applications/LBAF_app.py
marcinwrobel1986: Ok, that makes sense.

src/lbaf/Execution/lbsInformAndTransferAlgorithm.py
```diff
@@ -315,8 +308,8 @@ def execute(self, phase: Phase, distributions: dict, statistics: dict, a_min_max
             n_ignored, n_transfers, n_rejects = self.transfer_stage()
             n_proposed = n_transfers + n_rejects
             if n_proposed:
-                self.__logger.info(f"{n_proposed} proposed transfers, {n_transfers} occurred, {n_rejects} rejected "
-                                   f"({100. * n_rejects / n_proposed:.4}%)")
+                self.__logger.info(
+                    f"{n_proposed} proposed transfers, {n_transfers} occurred, {n_rejects} rejected ({100. * n_rejects / n_proposed:.4}%)")
```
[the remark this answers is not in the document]
ppebay: Is this a **requirement** as per language standard? 

I find splitting strings inelegant.

src/lbaf/IO/configurationValidator.py
```diff
@@ -102,35 +146,49 @@ def validate(valid_schema: Schema, schema_to_validate: dict):
     def main(self):
         """ Main routine for the config validation. """
         # Validate skeleton
-        if self.is_valid(valid_schema=self.__skeleton, schema_to_validate=self.__config_to_validate):
```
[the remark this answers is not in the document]
ppebay: Not easy to read. Easier to read/understand the function parameters when nicely aligned rather than in a single line.

src/lbaf/Applications/LBAF_app.py
marcinwrobel1986: `error message + exiting` can not be tested in unit test
Raising Error/Exception is exactly the same, but it can be tested

src/lbaf/IO/configurationValidator.py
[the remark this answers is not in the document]
ppebay: Yes but such a resolution is hard to attain on a laptop.
It is easier to read/understand the code when function parameters are aligned in consecutive lines than when crammed into a single line.

src/lbaf/IO/configurationValidator.py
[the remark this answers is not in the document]
ppebay: Same comment as above.

Also indent does not appear **wrong**: it does not result in a syntactic error.

src/lbaf/Execution/lbsInformAndTransferAlgorithm.py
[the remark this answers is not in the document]
marcinwrobel1986: It is usually depending on company standard 79-120 chars. 120 is very common and it fits 1280x.. resolution

src/lbaf/IO/configurationValidator.py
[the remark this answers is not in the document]
marcinwrobel1986: I think standard is 1920px, I have 2560px
Below you can not tell what's inside if, becasue indentation is wrong
```Python3
        if self.is_valid(
            valid_schema=self.__skeleton,
            schema_to_validate=self.__config_to_validate):
            self.__logger.info("Skeleton schema is valid")
```

src/lbaf/IO/configurationValidator.py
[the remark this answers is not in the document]
marcinwrobel1986: As I wrote above, Python interpreter can parse it, but it is not compliant with PEP8 and is a PEP8 violation.
`PEP 8: E125 continuation line with same indent as next logical line`

nlslatt:
It would be great to change the spread of objects to be based on the maximum number that actually hit a single rank as opposed to the total number of objects.

Also, it would be great to determine the random jitters in advance so that they can be consistent from phase to phase.

src/lbaf/IO/configurationValidator.py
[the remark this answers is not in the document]
ppebay: But what do you mean by **wrong**? Is it **wrong** in the sense that it causes a parsing/syntax error?

Or is **wrong** in a more relaxed sense? Like "it does not look good"?

src/lbaf/Execution/lbsInformAndTransferAlgorithm.py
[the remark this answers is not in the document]
ppebay: Yes but is this a **requirement**? I am really curious because oftentimes I realize that there are some obscure Python rules that people use, but that are not necessarily defined clearly anywhere.

src/lbaf/IO/configurationValidator.py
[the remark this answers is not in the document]
marcinwrobel1986: This causes linter error and violates PEP8, which we agreed to follow.
`PEP 8: E125 continuation line with same indent as next logical line`

The other thing is looks really bad, because all 3 lines are idented identical, which is not right visually + it's an anti-pattern, please see below from two different sources

Zrzut ekranu z 2022-05-12 08-05-56

Zrzut ekranu z 2022-05-12 08-09-44

src/lbaf/Execution/lbsInformAndTransferAlgorithm.py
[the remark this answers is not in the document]
marcinwrobel1986: Philippe if by requirement you mean something in the language or interpreter, then it is not. This is the best practice developed by years of working with the language by companies. I agree with that. I am not saying anything about rules defining how many variables can be there in a function, or how nested can you go with ifs and for loops. As I mention many times there are many usefull tools that have these rules integrated.

src/lbaf/IO/configurationValidator.py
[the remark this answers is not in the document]
ppebay: Thanks for the reference -- so let me make sure I get this right: the anti-pattern here is the fact that the conditional block is aligned with the condition itself, right?
```
if clause_1 or \
    clause_2:
    conditional_function()
```
That is the anti-pattern, right?

If so the PEP8 states that it should read instead:
```
if clause_1 or \
    clause_2:
        conditional_function()
```
Is that so?

We need to be completely clear about what we mean by **wrong**: does this cause the code to **break** or does it only violate some PEP8 style rule. I find it problematic with Python (unlike with other languages) that there is **syntactically correct** code that is nonetheless considered **incorrect** -- this opens the door to all sorts of interpretations/misunderstandings/disagreements about how the code should be.

If we want to say PEP8 is a strict requirement, I am not against it, but it also opens the door to us having to followed all sorts of extraneous rules that oftentimes seem more arbitrary/subjective than anything less. 

Do you see my point?

src/lbaf/Execution/lbsInformAndTransferAlgorithm.py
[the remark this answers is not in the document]
ppebay: I see your point but on the other hand I resist it because it seems more subjective than logically strong. In fact it makes us depend on some specific IDE implementations, or even to obscure rules that seem to come out of nowhere, even when we think (oftentimes for good reason) that **our** way of doing it actually reads better.

My general take on these things is that **if it is allowed syntactically by the language then it is correct**. This gives more freedom to the developer to express his intent when writing the code. But I realize there is a trade-off to be made here between the developer's viewpoint, and the general community's view on these things.

src/lbaf/IO/configurationValidator.py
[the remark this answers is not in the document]
ppebay: By the way I just noticed something else in the code here, which is a more fundamental problem:
```
        # Validate skeleton
        if self.is_valid(
            valid_schema=self.__skeleton,
            schema_to_validate=self.__config_to_validate):
                self.__logger.info("Skeleton schema is valid")
        else:
            self.__logger.error("Skeleton schema is invalid")
            self.validate(
                valid_schema=self.__skeleton,
                schema_to_validate=self.__config_to_validate)
```
In the `else` block, we have an error message that is not followed by a clean exit, but instead is followed by a subsequent call to the `validate()` method. This is against the practice throughout LBAF that is to `sys.exit(1)` whenever an error is encountered. 
So instead of calling `validate()` here the code should terminate in error. 

Or, do you want to rather consider this `else` clause as leading not to an **error**, but to a mere **warning**?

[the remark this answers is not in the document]
marcinwrobel1986: > Thanks for the reference -- so let me make sure I get this right: the anti-pattern here is the fact that the conditional block is aligned with the condition itself, right?
> 
> ```
> if clause_1 or \
>     clause_2:
>     conditional_function()
> ```
> 
> That is the anti-pattern, right?
> 
> If so the PEP8 states that it should read instead:
> 
> ```
> if clause_1 or \
>     clause_2:
>         conditional_function()
> ```
> 
> Is that so?

Yes, but the best would be, so we use line ot it's full length (120 chars):
```Python3
if clause_1 or clause_2:
    conditional_function()
```


> We need to be completely clear about what we mean by wrong: does this cause the code to break or does it only violate some PEP8 style rule. I find it problematic with Python (unlike with other languages) that there is syntactically correct code that is nonetheless considered incorrect -- this opens the door to all sorts of interpretations/misunderstandings/disagreements about how the code should be.

> If we want to say PEP8 is a strict requirement, I am not against it, but it also opens the door to us having to obey to all sorts of extraneous rules that oftentimes seem more arbitrary/subjective than anything less.

> Do you see my point?

I am pushing on the `minimum`, so the code is consistent and IMHO it is better to use the width of the screen and better to have screen covered with code instead of scrolling all the time.

src/lbaf/Execution/lbsInformAndTransferAlgorithm.py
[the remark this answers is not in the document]
marcinwrobel1986: > I see your point but on the other hand I resist it because it seems more subjective than logically strong. In fact it makes us depend on some specific IDE implementations, or even to obscure rules that seem to come out of nowhere, even when we think (oftentimes for good reason) that **our** way of doing it actually reads better.
> 
> My general take on these things is that **if it is allowed syntactically by the language then it is correct**. This gives more freedom to the developer to express his intent when writing the code. But I realize there is a trade-off to be made here between the developer's viewpoint, and the general community's view on these things.

I disagree with that. IDE has nothing to do with language rules. It is just the tool, that makes code more readable and help catching obvious mistakes. 

Regarding `if it is allowed syntactically by the language then it is correct` I disagree as well. If there are no rules/standards/best practices, then code goes momentally not readable, non maintaiable. Smart people and companies developed that for years, so if something is good and helps I will take it. Freedom is very important, but devs need to cooperate and write code which is understable for others. If I would be hit by a car today, then tomorrow you need somebody, who can understand what's going on.

src/lbaf/IO/configurationValidator.py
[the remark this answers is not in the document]
marcinwrobel1986: > By the way I just noticed something else in the code here, which is a more fundamental problem:
> 
> ```
>         # Validate skeleton
>         if self.is_valid(
>             valid_schema=self.__skeleton,
>             schema_to_validate=self.__config_to_validate):
>                 self.__logger.info("Skeleton schema is valid")
>         else:
>             self.__logger.error("Skeleton schema is invalid")
>             self.validate(
>                 valid_schema=self.__skeleton,
>                 schema_to_validate=self.__config_to_validate)
> ```
> 
> In the `else` block, we have an error message that is not followed by a clean exit, but instead is followed by a subsequent call to the `validate()` method. This is against the practice throughout LBAF that is to `sys.exit(1)` whenever an error is encountered. So instead of calling `validate()` here the code should terminate in error.
> 
> Or, do you want to rather consider this `else` clause as leading not to an **error**, but to a mere **warning**? I am unsure as to what your intent is, but my point is that if we want to let the code proceed then it cannot be an **error**.
> 
> BTW the same question applies to all four such conditionals that are following each other in the `main()` function.

Please take a look at the unit tests. When interpreter goes to `else`, we know that `schema` is invalid, because we checked `f self.is_valid(..)`, then in else it validates wrong schema with `validate()`, this raise `SchemaError` and returns a message what is wrong and exits as well. This SchemaError is then chatched in the tests and output is checked if it's that what we expect. So all good here.

ppebay:
Modified object placement strategy:

Screen Shot 2022-05-12 at 2 19 29 PM


@nlslatt @lifflander

src/lbaf/Execution/lbsInformAndTransferAlgorithm.py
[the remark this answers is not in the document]
ppebay: They might be smart -- but I am smarter.

src/lbaf/IO/configurationValidator.py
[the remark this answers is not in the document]
marcinwrobel1986: Philippe, we already spoke about that, I am writing unit tests, I can not test things properly without raising an error or exception, catching it and checking the error message. I thought we agreed on that I will be changing that respectively when writing unit tests. Please mind that raising error/exception is equal to sys.exit(1), but it can be tested.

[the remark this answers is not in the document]
ppebay: I don't remember we discussed that, but I believe it if you say so ;)

The problem I am seeing with this approach right now is that we have to be sure that the code actually calling the routine where the exception is raised, will indeed sys.exit(1) was expected. So it introduces the risk that it's actually not done in the main calling code.

I just want to make sure that each time we log an error, the code will exit. And here we have the log in one function, and the exit in another one. That's what I find a bit fragile. But if you're confident this will be maintained correctly, so be it!

src/lbaf/Execution/lbsInformAndTransferAlgorithm.py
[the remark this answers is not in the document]
ppebay: Alright, I have re-split the print line (with the small difference that you will see @marcinwrobel1986 that I have made **explicit concatenation** by adding the `+` sign). I find implicit concatenation dangerous because if part of the concatenated string is replaced by a variable string, as could happen when doing a global replacement, this would result in a syntax error (as implicit concatenation cannot be done for a string that is evaluated at runtime), e.g.:
```
>>> a="123" "456"
>>> print(a)
123456
```
but
```
>>> b="456"
>>> a = "123" b
  File "<stdin>", line 1
    a = "123" b
              ^
SyntaxError: invalid syntax
```

src/lbaf/IO/lbsMeshWriter.py
```diff
@@ -0,0 +1,308 @@
...
+                self.__logger.debug(f"Arranging a maximum of {n_o_per_dim} objects per dimension in {rank_dims}")
+                o_resolution = self.__grid_resolution / (n_o_per_dim + 1.)
+                rank_size = [n_o_per_dim if d in rank_dims else 1 for d in range(3)]
+                centering = [0.5 * o_resolution * (n_o_per_dim - 1.)
+                             if d in rank_dims else 0.0 for d in range(3)]
+                    
+                for i, o in enumerate(objects):
+                    # Insert point using offset and rank coordinates
+                    jitter = [
+                        (random.random() - 0.5) * o_resolution * self.__object_jitter
```
nlslatt, line 238: Can the `random.random()` part of the jitter be computed much earlier on and stored as a property of the object itself, independent of the rank on which the object exists?

nlslatt (commented):
This is a great tool! I don't see major issues remaining but, because of my minimal experience with Python, I wouldn't consider my review thorough.

src/lbaf/IO/configurationValidator.py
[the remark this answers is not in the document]
marcinwrobel1986: Philippe, you can try it by yourself, just make sure Schema is not right, and you will see how it works.
raising Error/Exception will always exit, the only case is when you use `try` and `except`, but that's not the case here.

src/lbaf/Execution/lbsInformAndTransferAlgorithm.py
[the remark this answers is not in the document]
marcinwrobel1986: This will not happend, might if you do that on purpose. Always use f-string for a variable inside strings:
This is wrong `b="456"`, should be `b = "456"`
```Python3
b = "456"
a = f"123{b}"
```

marcinwrobel1986:
@ppebay Unit tests still failing

src/lbaf/IO/configurationValidator.py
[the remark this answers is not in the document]
ppebay: My question was really the following:
- if an exception in raised in `function_1()` (in `file_1.py`)
- but `function_1()` is called by `function_2()` (in `file_2.py`), and `function_2()` is where the `sys.exit(1)` **if exception raised call to ** `function_1()`
==> then there is the risk that if someone modifies at a later time  `function_2()` (in `file_2.py`) and inadvertently removes the `sys.exit(1)` **if exception raised call to ** `function_1()`, then the desired behavior will be changed because an error condition will no longer occur.

Please pay attention to the intent I am trying to achieve here: what I am saying is that be dissociating the place where the exception is raised, from the place where the subsequent exit will occur, we are introducing a maintenance liability. 

My point is not to say that **currently* the behavior is not what we want. My point is that this separation between place of exception from place of exit is making the code harder to maintain.

Are you seeing what I am trying to explain/convey here? It is **not** about current behavior, it is about future risks because of separation of exception from exit.

src/lbaf/IO/lbsMeshWriter.py
ppebay: Done!

ppebay:
@nlslatt 2 identical frames at iterations 7 and 8 when steady-state was reached (per-object jitter is now invariant across entire simulation):
Screen Shot 2022-05-14 at 8 09 30 PM
Screen Shot 2022-05-14 at 8 09 41 PM

ppebay:
@marcinwrobel1986 I restored the unit test, it should work now.

Could we make sure that we no longer have a default `conf.yaml` file in Applications, but, instead, that the current `conf.yaml` becomes one of the canonical tests?

src/lbaf/IO/configurationValidator.py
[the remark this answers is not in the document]
marcinwrobel1986: I have checked the code a little bit on last Friday and there are places where we logging error and then sys.exit, but this error actually never happends, because in that case the error will be first noticed by ConfigurationValidator, this is something which needs to be cleaned up.
Regarding your comment from above:
We need to cover as much as we can  with unit tests, and in these unit tests we need to try to raise the errors/exceptions and see what is really reachable and remove all unnecessary code. Once we have that, it's very easy, one doesn't have to think what is intended, because it will be all covered in unit tests.
Modus operandi:
- one creates branch, one checks unit tests (unit tests should pass)
- one changes code
- one pushes a commit
- one checks unit tests and from there one sees what's expected and what's checked by the tests

Assuming, I don't like to speak what happend if, because anything could have happend and this discussion could be endless. I would like to really check the code, write unit tests and this really would be our guide. Any failing test would show to one, that something is wrong and should be changed or fixed.

P.S.
Regarding `Schema` library and this `SchemaError`, current behaviour in LBAF is due to this library and how it is designed. We could raise and error, just by checking if `schema.is_valid()`, but this bring us nothing, but when we then `schema.validate()` it will tell us where is the error.

marcinwrobel1986:
> @marcinwrobel1986 I restored the unit test, it should work now.

Still failing.

> Could we make sure that we no longer have a default `conf.yaml` file in Applications, but, instead, that the current `conf.yaml` becomes one of the canonical tests?

I think we need an issue, to remove this configuration YAML file from Applications, but we need to document is VERY WELL in wiki and readme file and always edit them if we change anything in config in future.

src/lbaf/Execution/lbsInformAndTransferAlgorithm.py
```diff
@@ -315,8 +308,10 @@ def execute(self, phase: Phase, distributions: dict, statistics: dict, a_min_max
             n_ignored, n_transfers, n_rejects = self.transfer_stage()
             n_proposed = n_transfers + n_rejects
             if n_proposed:
-                self.__logger.info(f"{n_proposed} proposed transfers, {n_transfers} occurred, {n_rejects} rejected "
-                                   f"({100. * n_rejects / n_proposed:.4}%)")
+                self.__logger.info(
```
marcinwrobel1986, line 311: @ppebay please make it as it was:

```python3
self.__logger.info(f"{n_proposed} proposed transfers, {n_transfers} occurred, {n_rejects} rejected "
                   f"({100. * n_rejects / n_proposed:.4}%)")
```

src/lbaf/Execution/lbsInformAndTransferAlgorithm.py
```diff
@@ -210,10 +203,11 @@ def transfer_stage(self):
             self.__logger.debug(f"Trying to offload from rank {p_src.get_id()} to {[p.get_id() for p in targets]}:")
 
             # Offload objects for as long as necessary and possible
-            srt_proc_obj = list(self.__order_strategy(p_src.get_migratable_objects(), p_src.get_id()))
-            while srt_proc_obj:
+            srt_rank_obj = list(self.__order_strategy(
```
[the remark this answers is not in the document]
marcinwrobel1986: BTW. Line 44 is too long, would be superb if you could change that as well, thanks!

src/lbaf/IO/configurationValidator.py
```diff
@@ -27,61 +33,99 @@ def __init__(self, config_to_validate: dict, logger: Logger):
...
+                error="Should be of type 'int' and > 0"),
+            Optional("generate_meshes"): {
+                "x_ranks": And(int, lambda x: x > 0,
+                               error="Should be of type 'int' and > 0"),
+                "y_ranks": And(int, lambda x: x > 0,
+                               error="Should be of type 'int' and > 0"),
+                "z_ranks": And(int, lambda x: x > 0,
+                               error="Should be of type 'int' and > 0"),
+                "object_jitter": And(float, lambda x: abs(x) < 1.0,
+                               error="Should be of type 'float' and magnitude < 1")
```
marcinwrobel1986, line 64: @ppebay indentation is wrong (PEP8), should be under float on line 63

I don't like the changes you made here, it is not readable for me.

src/lbaf/IO/configurationValidator.py
```diff
@@ -102,35 +146,49 @@ def validate(valid_schema: Schema, schema_to_validate: dict):
     def main(self):
         """ Main routine for the config validation. """
         # Validate skeleton
-        if self.is_valid(valid_schema=self.__skeleton, schema_to_validate=self.__config_to_validate):
+        if self.is_valid(
+            valid_schema=self.__skeleton,
+            schema_to_validate=self.__config_to_validate):
```
marcinwrobel1986, line 151: @ppebay PEP 8: E125 continuation line with same indent as next logical line

IS:
```Python3
        if self.is_valid(
            valid_schema=self.__skeleton,
            schema_to_validate=self.__config_to_validate):
            self.__logger.info("Skeleton schema is valid")
```

SHOULD BE:
```Python3
        if self.is_valid(
                valid_schema=self.__skeleton,
                schema_to_validate=self.__config_to_validate):
            self.__logger.info("Skeleton schema is valid")
```

src/lbaf/IO/configurationValidator.py
```diff
@@ -102,35 +146,49 @@ def validate(valid_schema: Schema, schema_to_validate: dict):
...
+                valid_schema=self.__skeleton,
+                schema_to_validate=self.__config_to_validate)
 
         # Validate from_data/from_samplers
         if (from_data := self.__config_to_validate.get("from_data")) is not None:
             self.__logger.info("Reading from data was chosen.")
-            if self.is_valid(valid_schema=self.__from_data, schema_to_validate=from_data):
+            if self.is_valid(
+                valid_schema=self.__from_data,
+                schema_to_validate=from_data):
```
marcinwrobel1986, line 164: @ppebay PEP 8: E125 continuation line with same indent as next logical line
SAME AS LINE 151

src/lbaf/IO/configurationValidator.py
```diff
@@ -102,35 +146,49 @@ def validate(valid_schema: Schema, schema_to_validate: dict):
...
-                self.validate(valid_schema=self.__from_data, schema_to_validate=from_data)
+                self.validate(
+                    valid_schema=self.__from_data,
+                    schema_to_validate=from_data)
         elif (from_samplers := self.__config_to_validate.get("from_samplers")) is not None:
             self.__logger.info("Simulate from samplers was chosen.")
-            if self.is_valid(valid_schema=self.__from_samplers, schema_to_validate=from_samplers):
+            if self.is_valid(
+                valid_schema=self.__from_samplers,
+                schema_to_validate=from_samplers):
```
marcinwrobel1986, line 175: @ppebay PEP 8: E125 continuation line with same indent as next logical line
SAME AS LINE 151

src/lbaf/IO/configurationValidator.py
```diff
@@ -102,35 +146,49 @@ def validate(valid_schema: Schema, schema_to_validate: dict):
...
+                    schema_to_validate=from_samplers)
 
         # Validate algorithm
         if (algorithm := self.__config_to_validate.get("algorithm").get("name")) is not None:
             self.__logger.info(f"Checking algorithm schema of: {algorithm}")
-            if self.is_valid(valid_schema=self.__algorithm.get(algorithm),
-                             schema_to_validate=self.__config_to_validate.get("algorithm")):
+            if self.is_valid(
+                valid_schema=self.__algorithm.get(algorithm),
+                schema_to_validate=self.__config_to_validate.get("algorithm")):
```
marcinwrobel1986, line 188: @ppebay PEP 8: E125 continuation line with same indent as next logical line
SAME AS LINE 151

src/lbaf/IO/lbsGridStreamer.py
```diff
@@ -9,9 +9,8 @@ class GridStreamer:
     """A class containing to stream a grid with time-varying attributes
     """
 
-    def __init__(self, points, lines, field_arrays, point_arrays, cell_arrays, lgr: Logger = None):
-        """Class constructor
-        """
+    def __init__(self, points: vtk.vtkPoints, lines: vtk.vtkCellArray, field_arrays: dict = {}, point_arrays: list = [], cell_arrays: list = [], lgr: Logger = None):
```
marcinwrobel1986, line 12: @ppebay line too long more than 120 chars
AND Default arguments value are mutable:
- field_arrays
- point_arrays
- cell_arrays

[Sample explanation](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/)

marcinwrobel1986: BTW line 49 is too long as well

src/lbaf/IO/lbsMeshWriter.py
```diff
@@ -0,0 +1,306 @@
...
+
+from .lbsGridStreamer import GridStreamer
+from ..Model.lbsPhase import Phase
+
+
+class MeshWriter:
+    """A class to write LBAF results to mesh files via VTK layer
+    """
+
+    def __init__(self, p: Phase, grid_size, object_jitter: float, f="lbs_out", r=1., output_dir=None, logger: Logger = None):
```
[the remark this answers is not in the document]
marcinwrobel1986: @ppebay

src/lbaf/IO/lbsMeshWriter.py
marcinwrobel1986: @ppebay

src/lbaf/IO/lbsMeshWriter.py
marcinwrobel1986: @ppebay

src/lbaf/IO/lbsMeshWriter.py
marcinwrobel1986: @ppebay

src/lbaf/IO/lbsMeshWriter.py
marcinwrobel1986: @ppebay

src/lbaf/IO/lbsMeshWriter.py
marcinwrobel1986: @ppebay

src/lbaf/IO/lbsMeshWriter.py
marcinwrobel1986: @ppebay

src/lbaf/IO/lbsMeshWriter.py
marcinwrobel1986: @ppebay

src/lbaf/Model/lbsPhase.py
marcinwrobel1986: @ppebay

marcinwrobel1986 (changes requested):
@ppebay I reviewed it once again and left 31 comments in places where cahnges needs to be done

ppebay:
@marcinwrobel1986 

What's the issue with that block in the mesh writer? 

```
            f"{grid_size[1]}x{grid_size[2]} rectilinear grid")

        # Ensure that a valid object jitter coefficient was passed
        if not isinstance(object_jitter, float) or abs(object_jitter) >= 1.0:
```

marcinwrobel1986:
> if not isinstance(object_jitter, float) or abs(object_jitter) >= 1.0:

This condition is checked in configuration validator. If it's invalid will be raised there (in configuration validator) and code will never go here, so it should be erased.

ppebay:
Alright so @marcinwrobel1986 do you mean that we can **confidently** remove all parameter type/range checks inside the code now?

marcinwrobel1986:
> @marcinwrobel1986
> 
> Actually no: I entered a negative value in the configuration file for the jitter:
> 
> ```
>   object_jitter: -0.5
> ```
> 
> and it is **not** caught by the configuration validator, hereby resulting in the following runtime error if I disable the `<= 0` sanity check in the mesh writer:
> 
> ```
> Traceback (most recent call last):
>   File "LBAF_app.py", line 326, in <module>
>     LBAFApp().main()
>   File "LBAF_app.py", line 269, in main
>     ex_writer = MeshWriter(
>   File "/Users/pppebay/Documents/Git/LB-analysis-framework/src/lbaf/IO/lbsMeshWriter.py", line 41, in __init__
>     f"Mapping {self.__n_p} ranks onto a {grid_size[0]}x"
> AttributeError: 'MeshWriter' object has no attribute '_MeshWriter__n_p'
> ```

@ppebay that means that, the condition you put in configuration validator is wrong, please see below.

Zrzut ekranu z 2022-05-20 12-31-40

marcinwrobel1986:
If negative value for `object_jitter` is not allowed it should be catched in configuration validator

ppebay:
@marcinwrobel1986 alright it's indeed catching for example a value of `10.5`:
```
During handling of the above exception, another exception occurred:

  File "LBAF_app.py", line 326, in <module>
    LBAFApp().main()
  File "LBAF_app.py", line 156, in __init__
    self.params = internalParameters(config_file=self.config_file)
  File "LBAF_app.py", line 63, in __init__
    self.configuration_validation()
  File "LBAF_app.py", line 87, in configuration_validation
    ConfigurationValidator(config_to_validate=self.configuration, logger=self.logger).main()
  File "/Users/pppebay/Documents/Git/LB-analysis-framework/src/lbaf/IO/configurationValidator.py", line 155, in main
    self.validate(
  File "/Users/pppebay/Documents/Git/LB-analysis-framework/src/lbaf/IO/configurationValidator.py", line 144, in validate
    return valid_schema.validate(schema_to_validate)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/schema.py", line 409, in validate
    raise SchemaError([message] + x.autos, [e.format(data) if e else None] + x.errors)
schema.SchemaError: Should be of type 'float' and magnitude < 1
```

That said we should not have this kind of runtime error. We should have a clean exit (which includes the SchemaError messate -- that part is fine -- but just not a crash like that). Can you please modify the exiting mechanism and make it a graceful failure? Thanks.

ppebay:
@marcinwrobel1986 By _"graceful failure"_ I mean something like that:
```
[lbsMeshWriter] Mapping 4 ranks onto a 2x2x1 rectilinear grid
# Object jitter should be of type 'float' and magnitude < 1

[pppebay@localhost]~/Documents/Git/LB-analysis-framework/src/lbaf/Applications$ echo $status
1
```

marcinwrobel1986:
> If negative value for `object_jitter` is not allowed it must be catched in configuration validator
> 
> @marcinwrobel1986 -- actually -0.5 is indeed correct :) I am indeed allowing for values in ]-1 ; 1[ in the random noise stuff: jittering on the negative side is also allowed (and is symmetrical).

If it's correct, then it shouldn't break the code.

marcinwrobel1986:
> @marcinwrobel1986 alright it's indeed catching for example a value of `10.5`:
> 
> ```
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "LBAF_app.py", line 326, in <module>
>     LBAFApp().main()
>   File "LBAF_app.py", line 156, in __init__
>     self.params = internalParameters(config_file=self.config_file)
>   File "LBAF_app.py", line 63, in __init__
>     self.configuration_validation()
>   File "LBAF_app.py", line 87, in configuration_validation
>     ConfigurationValidator(config_to_validate=self.configuration, logger=self.logger).main()
>   File "/Users/pppebay/Documents/Git/LB-analysis-framework/src/lbaf/IO/configurationValidator.py", line 155, in main
>     self.validate(
>   File "/Users/pppebay/Documents/Git/LB-analysis-framework/src/lbaf/IO/configurationValidator.py", line 144, in validate
>     return valid_schema.validate(schema_to_validate)
>   File "/opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/schema.py", line 409, in validate
>     raise SchemaError([message] + x.autos, [e.format(data) if e else None] + x.errors)
> schema.SchemaError: Should be of type 'float' and magnitude < 1
> ```
> 
> That said we should not have this kind of runtime error. We should have a clean exit (which includes the SchemaError messate -- that part is fine -- but just not a crash like that). Can you please modify the exiting mechanism and make it a graceful failure? Thanks.

@ppebay Philippe I told you, this is how this library `schema` works and how it's handling `SchemaErrors`

ppebay:
@marcinwrobel1986 I know you told me this is a behavior of that library but I think it is a bad behavior, (this is an unclean way to handle termination-on-error). Can't we fix it by catching the exception when invoking that library, and exiting cleanly? 
In other words, can't we encapsulate the call to that library inside a wrapper that would terminate cleanly?

We need to exit cleanly, with an error code that can be further used (as I have shown above with the `echo $status` returning the value of 1 in this instance.

marcinwrobel1986:
> @marcinwrobel1986 I know you told me that it is a behavior of that library -- but I think it is a bad behavior, (this is an unclean way to handle termination-on-error). Can't we fix it by catching the exception when invoking that library, and exiting cleanly? In other words, can't we encapsulate the call to that library inside a wrapper that would terminate cleanly?
> 
> We need to exit cleanly, with an error code that can be further used (as I have shown above with the `echo $status` returning the value of 1 in this instance.
> 
> That would make for a much nicer/cleaner way to interact with LBAF from other tools.

There are probably two ways:
- if one consider that as issue, one could write an issue [HERE](https://github.com/keleshev/schema/issues), but probably it will be left without answer
- as this is Python, a wrapper probably could be applied and overwrite the current behaviour, but once source code of the library (the new version) comes out it could break the behavior of the wrapper, which leave potential maintainability problem + time needed to prepare a wrapper, which could be use to do something else

ppebay:
@marcinwrobel1986 

I think this approach (and including the _caveat_ you mentioned) is what I have in mind:

> as this is Python, a wrapper probably could be applied and overwrite the current behaviour, but once source code of the library (the new version) comes out it could break the behavior of the wrapper, which leave potential maintainability problem + time needed to prepare a wrapper, which could be use to do something else

I will make an issue for that. I think it's worth investigating it.

ppebay:
@marcinwrobel1986 it seems that nothing is left pending. Am I leaving something aside?

marcinwrobel1986:
> @marcinwrobel1986 it seems that nothing is left pending. Am I leaving something aside?

@ppebay you asked me to do another review yesterday, so I did. there are a lot things to do. Please look at my comments from yesterday. I left yesterday 31 or 32 I believe.

ppebay:
@marcinwrobel1986 I did but everything seems resolved
Please answer my question precisely and tell me **what** is left. Thank you.

marcinwrobel1986:
> @marcinwrobel1986 I did but everything seems resolved (I am not seeing any unresolved conversation in the thread above: what am I missing?). Please answer my question precisely and tell me **what** is left. Thank you.

- data/visualization_examples/output_LBAF_view.py
Hardcoded paths to local files
Zrzut ekranu z 2022-05-20 17-35-22

- src/lbaf/Execution/lbsInformAndTransferAlgorithm.py
Zrzut ekranu z 2022-05-20 17-40-58

- src/lbaf/IO/configurationValidator.py
Zrzut ekranu z 2022-05-20 17-42-52
Zrzut ekranu z 2022-05-20 17-43-59
Zrzut ekranu z 2022-05-20 17-45-31
Zrzut ekranu z 2022-05-20 17-46-33
Zrzut ekranu z 2022-05-20 17-47-06
Zrzut ekranu z 2022-05-20 17-47-27

- src/lbaf/IO/lbsGridStreamer.py
Zrzut ekranu z 2022-05-20 17-48-12

- src/lbaf/IO/lbsMeshWriter.py
Zrzut ekranu z 2022-05-20 17-49-37
Zrzut ekranu z 2022-05-20 17-52-28
Zrzut ekranu z 2022-05-20 17-52-49

I think that is all.

ppebay:
@marcinwrobel1986 it seems you did not see my responses/commits. This is all addressed now.

marcinwrobel1986:
> @marcinwrobel1986 it seems you did not see sone of my responses/commits. This is all addressed now regarding the ParaView script: it's just an example and I moved it to the examples dir. Have you seen my response for that?

I have seen everything, I didn't mean to move it. It is ok if you want to move I don't mind, but there shouldn't be any paths which corresponds to your local machine like:
- /Users/pppebay/Documents/Git/LB-analysis-framework/output/output_file_object_view_00.vtp up to 08.vtp
- /Users/pppebay/Documents/Git/LB-analysis-framework/output/output_file_rank_view.e

ppebay (commented):
Parsed again all past review comments to make sure none was missed.

src/lbaf/Execution/lbsInformAndTransferAlgorithm.py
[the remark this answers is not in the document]
ppebay: I had missed that one!
Fixing it now @marcinwrobel1986

src/lbaf/IO/configurationValidator.py
ppebay: Missed that one too...

Should be good now. Tx.

src/lbaf/Execution/lbsInformAndTransferAlgorithm.py
ppebay: Should be done now but with **explicit** concatenation rather than **implicit**.

src/lbaf/IO/configurationValidator.py
ppebay: I think this one is now in the commit.

src/lbaf/IO/configurationValidator.py
ppebay: Should be fixed now

src/lbaf/IO/lbsGridStreamer.py
ppebay: That one was already there before that issue. Let's make it a separate issue -- I need to completely rework the grid streamer anyway.

src/lbaf/IO/configurationValidator.py
```diff
@@ -102,35 +146,49 @@ def validate(valid_schema: Schema, schema_to_validate: dict):
...
                 self.__logger.info("from_data schema is valid")
             else:
                 self.__logger.error("from_data schema is invalid")
-                self.validate(valid_schema=self.__from_data, schema_to_validate=from_data)
+                self.validate(
+                    valid_schema=self.__from_data,
+                    schema_to_validate=from_data)
         elif (from_samplers := self.__config_to_validate.get("from_samplers")) is not None:
             self.__logger.info("Simulate from samplers was chosen.")
-            if self.is_valid(valid_schema=self.__from_samplers, schema_to_validate=from_samplers):
```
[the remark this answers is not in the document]
ppebay: Already discussed in conversaton

src/lbaf/IO/configurationValidator.py
```diff
@@ -102,35 +146,49 @@ def validate(valid_schema: Schema, schema_to_validate: dict):
...
+                self.validate(
+                    valid_schema=self.__from_samplers,
+                    schema_to_validate=from_samplers)
 
         # Validate algorithm
         if (algorithm := self.__config_to_validate.get("algorithm").get("name")) is not None:
             self.__logger.info(f"Checking algorithm schema of: {algorithm}")
-            if self.is_valid(valid_schema=self.__algorithm.get(algorithm),
-                             schema_to_validate=self.__config_to_validate.get("algorithm")):
+            if self.is_valid(
```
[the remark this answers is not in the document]
ppebay: legibility

src/lbaf/IO/configurationValidator.py
ppebay: It is optional.

ppebay (commented):
Last pass: it seems that nothing is still pending, and yet the merge is still blocked. Not sure what is left.

src/lbaf/IO/lbsMeshWriter.py
ppebay: Should be all the same now in this file

src/lbaf/Applications/conf.yaml
ppebay: Ok restored to default

src/lbaf/IO/configurationValidator.py
ppebay: Extra lines are good and generally a way to enhance one's appreciation of life.

ppebay (commented):
extra pass

ppebay (commented):
reviewed

ppebay:
Re-requesting a last review @marcinwrobel1986 : despite the fact that all comments seem resolved, the merge is still blocked

marcinwrobel1986 (approved):
ok, we can merge this

PR #255, opened 2022-06-09
251 create phase stepper algorithm

ppebay:


ppebay:
This is WIP due to conflicts and difficulties processing data when rebasing with `develop`

ppebay:
You may want to try that PR with:
```
python LBAF_app.py --config step.yaml
```
from the `Applications` directory

marcinwrobel1986:
Hi Philippe @ppebay ,
please fix `Unit Tests` and `Acceptance Test` for Python 3.8 and I will take a look at the PR. Thanks

marcinwrobel1986:
> Yes, I think some of them were fixed by the latest commit, however do you know what this one is about @marcinwrobel1986 :
> 
> ```
>   ======================================================================
>   FAIL: test_config_validator_wrong_from_data_phase_name (test_configuration_validator.TestConfig)
>   ----------------------------------------------------------------------
>   Traceback (most recent call last):
>     File "/home/runner/work/LB-analysis-framework/LB-analysis-framework/tests/test_configuration_validator.py", line 89, in test_config_validator_wrong_from_data_phase_name
>       ConfigurationValidator(config_to_validate=configuration, logger=logger()).main()
>   AssertionError: SchemaMissingKeyError not raised
> ```
> 
> The error messages of the validator are cryptic, it would be nice to make them clearer than that.

Yes, this test was supposed to catch that in given configuration file there were no `phase_id` given, so the test expected `SchemaMissingKeyError` to be raised with message "Missing key: 'phase_id'"

ppebay:
Sure thing, so now we need to modify it to catch the following schema:
```
phase_ids:
  - <index_1>
  ...
  - <index_n>
  ```
that is now replacing the previous one:
```
phase_id: <unique_index>
```
? @marcinwrobel1986

**Comment by ppebay** [comment]
@marcinwrobel1986 I think most of the unit tests are fixed; there's one that does not appear to be related to the change from `phase_id` to `phased_ids`, can you please look into it?
```
Error: ERROR: test_lbs_object (unittest.loader._FailedTest)
  ----------------------------------------------------------------------
  ImportError: Failed to import test module: test_lbs_object
    File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/unittest/loader.py", line 436, in _find_test_path
      module = self._get_module_from_name(name)
    File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/unittest/loader.py", line 377, in _get_module_from_name
      __import__(name)
    File "/home/runner/work/LB-analysis-framework/LB-analysis-framework/tests/test_lbs_object.py", line 86
      self.assertEqual(err.exception.args[0], f"i: 25 is type of <class "str">! Must be <class "int">!")
```
Other than this the code should be ready to go.

**Comment by marcinwrobel1986** [comment]
> test_lbs_object

Hi Philippe @ppebay ,

the reason is very simple.
You do remember how you like `"` instead of `'` and I am using rather `'` instead of `"`, so now you will get a clear picture why.
I will explain a few cases for you.

You changed that in your [commit](https://github.com/DARMA-tasking/LB-analysis-framework/pull/255/commits/8922b514be94c3cf40f2886887cbaeb277cd91c7)

Please take a look at this fstring:
```Python3
# WAS:
    Object(i='25', t=2.5)
self.assertEqual(err.exception.args[0], f"i: 25 is type of <class 'str'>! Must be <class 'int'>!")

# After your change
    Object(i="25", t=2.5)
self.assertEqual(err.exception.args[0], f"i: 25 is type of <class "str">! Must be <class "int">!")
```
As you can see this is even interpreted differently in this block code on GitHub. After your changes str and int are treated as variables and there are actually three strings:
```Python3
# First one:
f"i: 25 is type of <class "

# Second one:
">! Must be <class "

# Third one:
">!"
```

The other thing is that, your change was actually wrong, because `f"i: 25 is type of <class 'str'>! Must be <class 'int'>!"` is actually a string which is returned by Python itself and there are single quotes `'` in that message, so even if you use escape characters before the quotes this will be wrong.
Normally I use single quotes for fstring, but in this case I used double, because the string inside had single quotes and I didn't want to escape that.

The other reason that I normally prefer to use single quotes is when one is working with HTML or JSON, there are double quotes used in HTML and JSON, so when I use single quotes at the beggining and at the end I don't have to escape each double quotes. The example:

Let's use this JSON, we know:
```JSON
{"type":"LBDatafile","phases":[]}
```
In order to assign this string to a variable I would use single quotes like that:
```Python3
json_as_str = '{"type":"LBDatafile","phases":[]}'
```
If I would use double quotes like that:
```Python3
json_as_str = "{"type":"LBDatafile","phases":[]}"
```
I would get a SyntaxError:
Zrzut ekranu z 2022-06-11 11-08-55

So in order, to do that right with double quotes I would need to escape every double quote inside double quotes like that:
```Python3
json_as_str = "{\"type\":\"LBDatafile\",\"phases\":[]}"
```
Which is very efficient, ambitious and every programmer in the world is eager to do that, rather than use single quotes.
I hope you know my reasons better now.
Back to your comment, this is very much related with your changes and must be fixed.

ppebay:
Oh sorry that must be a regexp-global-replace on my part that went awry. I will return the simple quotes there. My bad!

In general in the main code I am trying to keep simple quotes for single characters, as an analogy to C conventions, which makes it easier to translate code when necessary. It's just a style convention at this point, I would not be against changing it but we would have to change it everywhere for style consistency. Therefore for now I think we can keep the current convention, except when it's absolutely necessary to use simple quotes as in the CI/testing programs as you mentioned.

ppebay:
@marcinwrobel1986 @lifflander Should be ready now

marcinwrobel1986:
> @marcinwrobel1986 @lifflander Should be ready now

@ppebay Not really, you change something with imbalance file and it can not be found and checked. Please take a look on the Acceptance Tests Python 3.8 Thanks! Once that is fixed I will take a look tomorrow morning.

ppebay:
@marcinwrobel1986 I have no idea how this is setup in CI. 
When running it on my machine it passes and generates the correct `imbalance.txt` file: 
```
[lbsStatistics] Final imbalance: 0.25 logged to /Users/pppebay/Documents/Git/LB-analysis-framework/output/imbalance.txt
```
Can you please check what's going on in CI? I don't think it's a good use of my time to figure out the CI details. Thanks

marcinwrobel1986:
> @marcinwrobel1986 I have no idea how this is setup in CI. When running it on my machine it passes and generates the correct `imbalance.txt` file:
> 
> ```
> [lbsStatistics] Final imbalance: 0.25 logged to /Users/pppebay/Documents/Git/LB-analysis-framework/output/imbalance.txt
> ```
> 
> Can you please check what's going on in CI? I don't think it's a good use of my time to figure out the CI details. Thanks

@ppebay Philippe this is probably this [commit](https://github.com/DARMA-tasking/LB-analysis-framework/pull/255/commits/eb01e6d68c6131a7f22fea98cafff2d4c8ea2e5c)
I can check tomorrow, the thing is that file is probably not saved where it was.

ppebay:
@marcinwrobel1986 well there was apparently the same problem in CI before I made that change.

BTW, I have replaced `file` with `file_name` for good form, to avoid confusion between the file object and the string.

marcinwrobel1986:
> @marcinwrobel1986 well there was apparently the same problem in CI before I made that change.
> 
> BTW, I have replaced `file` with `file_name` for good form, to avoid confusion between the file object and the string.

Hello @ppebay 
I have found the problem:

Zrzut ekranu z 2022-06-13 06-29-58
Zrzut ekranu z 2022-06-13 06-30-11

You have changed the path in config file for CI.
It caused that, the imbalance file was saved in:
Zrzut ekranu z 2022-06-13 06-34-05
Which was totally unexpected.
I am pushing a commit which will change that to previous state.

src/lbaf/Execution/lbsPhaseStepperAlgorithm.py
```diff
@@ -0,0 +1,56 @@
+import sys
```
marcinwrobel1986, line 1: import not used, should be removed

src/lbaf/Execution/lbsPhaseStepperAlgorithm.py
```diff
@@ -0,0 +1,56 @@
...
+import math
```
marcinwrobel1986, line 2: import not used, should be removed

src/lbaf/Execution/lbsPhaseStepperAlgorithm.py
```diff
@@ -0,0 +1,56 @@
...
+import itertools
```
marcinwrobel1986, line 3: import not used, should be removed

src/lbaf/Execution/lbsPhaseStepperAlgorithm.py
```diff
@@ -0,0 +1,56 @@
...
+from logging import Logger
+
+from .lbsAlgorithmBase import AlgorithmBase
+from ..Model.lbsObjectCommunicator import ObjectCommunicator
```
marcinwrobel1986, line 7: import not used, should be removed

src/lbaf/Execution/lbsPhaseStepperAlgorithm.py
```diff
@@ -0,0 +1,56 @@
...
+
+        # Assign logger to instance variable
+        self.__logger = lgr
+
+    def execute(self, phases: list, distributions: dict, statistics: dict, _):
+        """ Execute brute force optimization algorithm on Phase instance
+        """
+        # Ensure that a list with at least one phase was provided
+        if not phases or not isinstance(phases, list) or not all(
+            [isinstance(p, Phase) for p in phases]):
```
marcinwrobel1986, line 32: should be in line 31. Then line 31 will have 106 chars < 120, so it's ok, more readable

src/lbaf/Execution/lbsBruteForceAlgorithm.py
```diff
@@ -63,12 +63,12 @@ def compute_arrangement_works(self, objects: tuple, arrangement: tuple) -> dict:
...
-    def execute(self, phase: Phase, distributions: dict, statistics: dict, _):
+    def execute(self, phases: list, distributions: dict, statistics: dict, _):
         """ Execute brute force optimization algorithm on Phase instance
         """
-
-        # Ensure that a phase was properly passed
-        if not isinstance(phase, Phase):
+        # Ensure that a list with at least one phase was provided
+        if not phases or not isinstance(phases, list) or not isinstance(
+            (phase := phases[0]), Phase):
```
marcinwrobel1986, line 71: should be in line 70. Then line 70 will have 102 chars < 120, so it's ok, more readable

marcinwrobel1986 (changes requested):
Just a few small changes.
I have run the `step.yaml` configuration.
Is imbalance supposed to be 1.0 from our `synthetic_lb_data` data?

src/lbaf/Execution/lbsBruteForceAlgorithm.py
ppebay: I like it better like it is now

src/lbaf/Execution/lbsPhaseStepperAlgorithm.py
ppebay: I like better to see all conditions at a glance on a laptop

src/lbaf/Execution/lbsPhaseStepperAlgorithm.py
ppebay: Done

src/lbaf/Execution/lbsPhaseStepperAlgorithm.py
ppebay: Done

ppebay (commented):
All changes seem to be in but it seems to still be blocking, do you know what I am missing @marcinwrobel1986 ?

ppebay:
@marcinwrobel1986 it's telling me there still is one change requested -- however they all appear resolved on my side.

Do you know which one that would be?

Thanks

marcinwrobel1986:
> @marcinwrobel1986 it's telling me there still is one change requested -- however they all appear resolved on my side.
> 
> Do you know which one that would be?
> 
> Thanks

Let me check that later today.

src/lbaf/Execution/lbsPhaseStepperAlgorithm.py
marcinwrobel1986: Then it is again the same problem. Indentation is wrong, why? 
PEP 8: E125 continuation line with same indent as next logical line
Because the indentation should be different than lines inside if:
WRONG:
```Python3
if not phases or not isinstance(phases, list) or not all(
    [isinstance(p, Phase) for p in phases]):
    self.__logger.error(f"Algorithm execution requires a Phase instance")
    raise SystemExit(1)
```
CORRECT:
```Python3
if not phases or not isinstance(phases, list) or not all(
        [isinstance(p, Phase) for p in phases]):
    self.__logger.error(f"Algorithm execution requires a Phase instance")
    raise SystemExit(1)
```

src/lbaf/Execution/lbsBruteForceAlgorithm.py
marcinwrobel1986: Then it is again the same problem. Indentation is wrong, why? 
PEP 8: E125 continuation line with same indent as next logical line
Because the indentation should be different than lines inside if:
WRONG:
```Python3
if not phases or not isinstance(phases, list) or not isinstance(
    (phase := phases[0]), Phase):
    self.__logger.error(f"Algorithm execution requires a Phase instance")
    raise SystemExit(1)
self.phase = phase
```

CORRECT:
```Python3
if not phases or not isinstance(phases, list) or not isinstance(
        (phase := phases[0]), Phase):
    self.__logger.error(f"Algorithm execution requires a Phase instance")
    raise SystemExit(1)
self.phase = phase
```

marcinwrobel1986 (changes requested):
Two more issues, regarding PEP 8: E125 continuation line with same indent as next logical line
Apart from that, it's ok

@marcinwrobel1986 please please for things like this don't hesitate to fix them on-the-fly, saves me a lot of time. Thanks!

src/lbaf/Execution/lbsBruteForceAlgorithm.py
ppebay: I don't know but that's a detail; please let's not waste our time with small things like this
We can always do a global style sweep later
We have bigger fish to fry now

marcinwrobel1986:
> @marcinwrobel1986 please please for things like this don't hesitate to fix them on-the-fly, saves me a lot of time. Thanks!

I wanted to point that out, because this is not the first time I mentioned that. It was in the previous PR's as well.

Regarding 
```
I don't know but that's a detail; please let's not waste our time with small things like this
We can always do a global style sweep later
We have bigger fish to fry now
```

I pointed that out, because it is a PEP8 violation `PEP 8: E125 continuation line with same indent as next logical line`. We agreed to follow PEP8 and I think it is important to follow.
If you ask me what I find not important is e.g. [this commit](https://github.com/DARMA-tasking/LB-analysis-framework/pull/255/commits/eff91aaec3c87ce8be61f03b258ebad8402d4790) or not important and danger as it appeared was [this commit](https://github.com/DARMA-tasking/LB-analysis-framework/pull/255/commits/8922b514be94c3cf40f2886887cbaeb277cd91c7)

I pushed the chages you wanted `please please for things like this don't hesitate to fix them on-the-fly, saves me a lot of time. Thanks!` and approved this PR.

marcinwrobel1986 (approved):
Should be good now

Issue #260, opened 2022-06-16
Add option to stabilize jitter in stepper algorithm when objects are consistent across phases


nlslatt:
The stepper visualizations exhibit a lot of unnecessary jitter when the objects visualized are consistent from one phase to the next, making it difficult to tell if objects are migrating. Any object ID found in one phase can be assumed to be the same object as one with an identical ID in a different phase.

When new objects are not being created in subsequent phases, remove the variable jitter (using a configuration option to enable this if needed).

If it's possible to add previously unseen objects to the end of a rank and keep everything else stable, that would be even better.

ppebay:
@nlslatt upon closer inspection, I realized that this is not a question of object jitter (which **is** indeed constant per object ID across all iterations and/or phases), but of ordering of sentinel vs. migratable objects: while the former remain (by definition) on the same rank across the entire run, the former (can) move. As a result, the per-rank local grids (a.k.a. "mesh blocks") are recomputed over a potentially different set of objects for each frame. 

At this point my thinking is that a simple ordering of objects (first the sentinels; then the migratable ones) should resolve the issue regarding the "flickering" of sentinel objects between frames. More later.

ppebay:
Resolved by #265

PR #265, opened 2022-06-17
#260: implemented issue (but it has nothing to do with jitter finally)

ppebay:
This had to do finally more with using consistent ordering across phases: sentinel objects first, then migratable ones -- and always ordering them inside their local "blocks".


nlslatt (approved):
The output now looks great!

PR #271, opened 2022-06-23
#263: two modes or erroring out in lbaf

### LINKED ISSUE #263: 2 modes or erroring-out in LBAF
When we have something that is incorrect, like a setting, an inconsistency in the data, etc. the desired behavior is indeed to have a “clean” termination with a clear, crisp one-liner telling what’s incorrect.  We can consider this the user-level erroring-out.

In contrast, when we have what I would call a “developer error”, we still want to full stack trace; it’s a different use case. However now we have something that prevents the stack tracing both for the “developer errors” and for when we are exiting in a pre-recognized error mode. That is not the desired behavior. 

Here is an example; there is obviously a syntax error some place in the code, but all we get is this error message:
```
[LBAF_app] NameError name 'o' is not defined
```
So in that case we do not want to block the brutal crash + stack dump; in fact we want them
Fundamentally, a distinction is to be made between 2 **different** types of errors:

1- One is an error where it corresponds to a previously identified error condition. Here is an example:
```
        # Make sure that Phase instances were passed
        if not all([isinstance(p, Phase) for p in phases]):
            self.__logger.error(
                "Mesh writer expects a list of Phase instances as input")
            raise SystemExit(1)
```
We know in advance what we want to verify, it’s a previously identified error mode. For that one, because we already know what it is, a clean termination, with a one-liner, is sufficient. It is telling the user, for instance, that the method was not given the expected input.It is necessary because Python’s not being typed.

2- The other error is, say, you have something like that in the code:
```
        bal;cl =- 23
```
This is an unidentified developer error. In that instance a syntax error. It could also be an error where we are erroneously referring to a member variable that does not exist:
```
       blah.o = 15
```
Where instance `blah` actually does not have a member `o`.
In the current setting that you’ve done, if I have the terse reporting flag, I will not get the full stack trace in the second case. But I want it. Because it is a developer error and we need full information about it, as it was not recognized before.

So, what I am saying here is that the “switch” approach does not work, because we want at the same time terse reporting for errors of type 1 (which follow a well-defined conditional) and full stack tracking for errors of type 2 (because they are not defined errors).
That is the desired behavior.

marcinwrobel1986:
### THIS PR:
- added exception handler to utils
- applied exception handler policy to all known error cases

Closes #263

### DIFF
--- a/src/lbaf/Applications/conf.yaml
+++ b/src/lbaf/Applications/conf.yaml
@@ -71,7 +71,7 @@ algorithm:
 
 # Specify output
 #logging_level: debug
-show_traceback: False
+show_traceback: True
 terminal_background: light
 generate_multimedia: False
 output_dir: ../../../output
--- a/src/lbaf/Applications/conf.yaml
+++ b/src/lbaf/Applications/conf.yaml
@@ -20,6 +20,7 @@
 #        skip_transfer [bool]  (default: False) skip transfer phase
 #      PhaseStepper
 # logging_level [str]          set to `info`, `debug`, `warning` or `error`
+# log_to_file [str]            filepath to save the log file (optional)
 # x_procs [int]                number of procs in x direction for rank visualization
 # y_procs [int]                number of procs in y direction for rank visualization
 # z_procs [int]                number of procs in z direction for rank visualization
--- a/src/lbaf/IO/lbsVTDataReader.py
+++ b/src/lbaf/IO/lbsVTDataReader.py
@@ -9,6 +9,7 @@ from .schemaValidator import SchemaValidator
 from ..Model.lbsObject import Object
 from ..Model.lbsObjectCommunicator import ObjectCommunicator
 from ..Model.lbsRank import Rank
+from ..Utils.exception_handler import exc_handler
 
 
 class LoadReader:
@@ -51,6 +52,7 @@ class LoadReader:
         file_name = self.get_node_trace_file_name(node_id)
         self.__logger.info(f"Reading {file_name} VT object map")
         if not os.path.isfile(file_name):
+            sys.excepthook = exc_handler
             raise FileNotFoundError(f"File {file_name} not found!")
 
         # Retrieve communications from JSON reader
@@ -87,6 +89,7 @@ class LoadReader:
             except KeyError as e:
                 msg_err = f"Could not retrieve information for rank {p} at time_step {phase_id}. KeyError {e}"
                 self.__logger.error(msg_err)
+                sys.excepthook = exc_handler
                 raise KeyError(msg_err)
 
             # Merge rank communication with existing ones
@@ -136,6 +139,7 @@ class LoadReader:
         # Extracting type from JSON data
         schema_type = decompressed_dict.get("type")
         if schema_type is None:
+            sys.excepthook = exc_handler
             raise TypeError("JSON data is missing 'type' key")
 
         # Validate schema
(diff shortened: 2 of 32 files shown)


src/lbaf/Applications/conf.yaml
```diff
@@ -72,7 +72,7 @@ algorithm:
 
 # Specify output
 #logging_level: debug
-show_traceback: False
+show_traceback: True
```
ppebay, line 75: Please eliminate this option, we don't want this any longer.  We will always have the 2 following modes:
1- pre-defined error condition: `logger.error` + exception --> termination without stack trace
2- undefined error: runtime error + stack trace

Therefore we no longer need this `show_traceback` error

src/lbaf/IO/lbsVTDataReader.py
```diff
@@ -51,6 +52,7 @@ def read(self, node_id: int, phase_id: int = -1, comm: bool = False) -> tuple:
         file_name = self.get_node_trace_file_name(node_id)
         self.__logger.info(f"Reading {file_name} VT object map")
         if not os.path.isfile(file_name):
+            sys.excepthook = exc_handler
             raise FileNotFoundError(f"File {file_name} not found!")
```
ppebay, line 56: No exclamation marks plz

src/lbaf/Utils/JSON_data_files_validator.py
```diff
@@ -60,6 +61,7 @@ def __get_files_for_validation(dir_path: str, file_prefix: str, file_suffix: str
         list_of_files = os.listdir(dir_path)
 
         if not list_of_files:
+            sys.excepthook = exc_handler
             raise FileNotFoundError(f"Directory: {dir_path} is EMPTY!")
```
ppebay, line 65: no excl mark

src/lbaf/Utils/data_stat_files_updater.py
```diff
@@ -121,8 +125,10 @@ def main(self):
                 for file in list_of_files_for_validation:
                     self.__add_type_to_file(file_path=file)
             else:
+                sys.excepthook = exc_handler
                 raise FileNotFoundError(f"Directory: {self.__dir_path} does NOT exist!")
         else:
+            sys.excepthook = exc_handler
             raise Exception("FILE path or DIRECTORY path has to be given!")
```
ppebay, line 132: same

src/lbaf/Utils/csv_2_json_data_converter.py
```diff
@@ -51,6 +53,7 @@ def _get_data_dir(dir_path: str) -> str:
             return os.path.join(project_path, dir_path)
         else:
             print(f"Can not find dir {dir_path}!")
```
ppebay, line 55: !

ppebay, line 55: no !

ppebay (changes requested):
Looking good for the most part, just a handful of changes

src/lbaf/Applications/conf.yaml
marcinwrobel1986: @ppebay  For that I created an [issue](https://github.com/DARMA-tasking/LB-analysis-framework/issues/272), to do when I will have some capacity. Rest is done.

ppebay: Okay got it. So we leave it like that for now.

ppebay (approved):

PR #278, opened 2022-08-22
#277: Accept range in phase indices in lbaf config

### LINKED ISSUE #277: Accept range in phase indices in lbaf config
e.g. if we want to simulate 100 phases we would want to be able to have something like:

```yml
from_data:
  phase_ids:
    - [0,100]
```

@ppebay mentioned I should file an issue for this

marcinwrobel1986:
### THIS PR:
- added info in conf.yaml
- added tests of enhanced configuration option with two sample configurations
- modified parsing data parameters in LBAF_app
- modified configurationValidator.py, so it accepts str and validates it with Regex

closes #277

### DIFF
--- a/src/lbaf/Applications/LBAF_app.py
+++ b/src/lbaf/Applications/LBAF_app.py
@@ -112,7 +112,10 @@ class internalParameters:
         # Parse data parameters if present
         if self.configuration.get("from_data") is not None:
             self.data_stem = self.configuration.get("from_data").get("data_stem")
-            self.phase_ids = self.configuration.get("from_data").get("phase_ids")
+            if isinstance(self.configuration.get("from_data").get("phase_ids"), str):
+                self.phase_ids = list(map(int, self.configuration.get("from_data").get("phase_ids").split('-')))
+            else:
+                self.phase_ids = self.configuration.get("from_data").get("phase_ids")
 
         # Parse sampling parameters if present
         if self.configuration.get("from_samplers") is not None:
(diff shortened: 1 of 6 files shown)


src/lbaf/Applications/LBAF_app.py
```diff
@@ -112,7 +112,10 @@ def parse_conf_file(self):
         # Parse data parameters if present
         if self.configuration.get("from_data") is not None:
             self.data_stem = self.configuration.get("from_data").get("data_stem")
-            self.phase_ids = self.configuration.get("from_data").get("phase_ids")
+            if isinstance(self.configuration.get("from_data").get("phase_ids"), str):
+                self.phase_ids = list(map(int, self.configuration.get("from_data").get("phase_ids").split('-')))
```
nmm0, line 116: Not super familiar with the LBAF code but should this be
```suggestion
                self.phase_ids = range(*map(int, self.configuration.get("from_data").get("phase_ids").split('-')))
```
It looks to me like this will just add the first and last index to the iterations rather than all the indices from the first to the last

marcinwrobel1986: @nmm0 Wow, I am really sorry! Good there is somebody checking :) - thanks a lot. Not exactly like that, but thanks for the tip, I did too many things at once today :) - sorry! changing to WIP, will be done latest tomorrow!

marcinwrobel1986: @nmm0 done. Thanks once more! Total blackout today :)

ppebay (approved):
Looking good

PR #288, opened 2022-09-08
#287: allow for a default 0-value of the jitter

### LINKED ISSUE #287: Allow for additional jitter values at later phases
This is to visualize phases with variable numbers of objects

ppebay:
fixes #287 

This is only a short-term fix, allowing for the visualization of data sets with variable number of objects across phases. It will only make sense if the jitter value is set to 0 globally.


nmm0 (approved):
This fixes the issue for me -- we should definitely add a random jitter if it doesn't exist next though

PR #299, opened 2022-11-02
#298: add json memory usage exemplar

### LINKED ISSUE #298: Add example of user-defined memory usage JSON fields
Upload a set of JSON files including memory usage data in the `"user-defined"` section. The fields contained will be:

- `"task_serialized_bytes"`: This is the serialized size of the task, which can be
  used for modeling the migration cost of the task. It should not be included
  when computing the memory usage on a rank.
- `"shared_id"`: This uniquely identifies a block of data on which multiple tasks
  will operate.
- `"shared_bytes"`: This is the size of the block of data being operated on by the
  relevant set of tasks. This memory cost will be incurred exactly once on each
  MPI rank on which a task with this `shared_id` exists.
- `"task_footprint_bytes"`: This is the footprinted size of the task in its
  non-running state. We will incur this memory cost once for each individual
  task, even if there are other tasks on this rank with the same `shared_id`. This
  can be greater than `task_serialized_bytes` when the task has data members that
  have greater capacity than is being used at serialization time.
- `"task_working_bytes"`: This is the high water mark of the additional working
  memory required by the individual task, such as temporary memory needed for
  intermediate computation. This value does not include memory shared with other
  tasks (i.e., `shared_bytes`), nor does it include the `task_footprint bytes` or
  `task_serialized_bytes`. This cost is incurred for each individual task, but
  only one at a time because tasks will not run concurrently.
- `"rank_working_bytes"`: This is the amount of memory that the particular rank
  needs while processing tasks. This may include global data, constants, and
  completely unrelated data pre-allocated by the application. It is assumed to
  be constant over time but may vary from rank to rank. This value does not
  include `shared_bytes`, `task_working_bytes`, `task_footprint_bytes`, or
  `task_serialized_bytes`.

The maximum memory usage for determining if task placement is feasible will be:
  `max_memory_usage = rank_working_bytes + shared_level_memory + max_task_level_memory`

Computing `shared_level_memory`: Let `S` be the set of unique `shared_id` values on
the rank being considered. Then `shared_level_memory` is simply the sum of
`shared_bytes` values for each `shared_id` in `S`.

Computing `max_task_level_memory`: Let `T` be the set of all tasks on a rank,
regardless of the `shared_id` on which they operate. Then `max_task_level_memory`
is the sum of `task_footprint_bytes` values for each task in `T` plus the maximum
over the `task_working_bytes` values for each task in `T`.

nlslatt:
Closes #298


marcinwrobel1986 (changes requested):
Hello @nlslatt ,
I have found one problem.
Mainly we are checking for `LBDatafile` or `LBStatsfile` type.
The one in your data is `LBDataFile` (with capital `F`).
The code I mentioned is in VT - in: `scripts/JSON_data_files_validator.py`

The question is what do we do with that?
1. We keep `LBDatafile` or `LBStatsfile`:
- this means you have to make changes in VT if it's already implemented
2. We keep `LBDataFile` or `LBStatsFile`:
- this means somebody has to change the types in VT
- I have to make sure our tests are passing in LBAF

Please let me know what have you decided. Thanks.

nlslatt:
@marcinwrobel1986 Thanks for catching that. I'm only indirectly using VT for outputting the files, so the typo was in my driver. I will fix this PR today.

PR #301, opened 2022-11-02
300 Add a toy problem for load balancing under memory constraints

### LINKED ISSUE #300: Add a toy problem for load balancing under memory constraints
Provide a set of JSON files containing memory usage fields that can be used to gauge whether a memory-aware load balancing algorithm is achieving a sensible solution.

nlslatt:
Closes #300 

See README for details.

For simplicity, the memory usage of tasks, blocks, and ranks don't vary. The goal here is to stay within the memory constrain while balancing the load, without breaking locality too badly (introducing unwarranted levels of communication).


marcinwrobel1986 (changes requested):
Same as in #299 

Hello @nlslatt ,
I have found one problem.
Mainly we are checking for `LBDatafile` or `LBStatsfile` type.
The one in your data is `LBDataFile` (with capital `F`).
The code I mentioned is in VT - in: `scripts/JSON_data_files_validator.py`

The question is what do we do with that?
1. We keep `LBDatafile` or `LBStatsfile`:
- this means you have to make changes in VT if it's already implemented
2. We keep `LBDataFile` or `LBStatsFile`:
- this means somebody has to change the types in VT
- I have to make sure our tests are passing in LBAF

Please let me know what have you decided. Thanks.

nlslatt:
@marcinwrobel1986 Thanks for catching that. I'm only indirectly using VT for outputting the files, so the typo was in my driver. I will fix this PR today.

marcinwrobel1986 (approved):
@nlslatt This one seems valid as well, thanks!

lifflander (approved):

PR #318, opened 2022-12-05
#317: Reformat style of lbsStatistics and remove special handling of text string for imbalance file output

### LINKED ISSUE #317: Reformat style of lbsStatistics and remove special handling of text string for imbalance file output
* Doc strings are not consistent with code style
* This special case:  `if var_name == "final rank load" and file_name is not None` should be detected and treated outside of the statistics printout routine itself.

ppebay:
Fixes #317


ppebay:
Recreating new PR to allow for clean rebase

PR #319, opened 2022-12-06
#317: Reformat style of lbsStatistics, remove special handling of text string for imbalance file

### LINKED ISSUE #317: Reformat style of lbsStatistics and remove special handling of text string for imbalance file output

ppebay:
Fixes #317

### DIFF
--- a/src/lbaf/Applications/LBAF_app.py
+++ b/src/lbaf/Applications/LBAF_app.py
@@ -400,15 +400,17 @@ class LBAFApp:
 
         # Compute and print final rank load and edge volume statistics
         curr_phase = phases[-1]
-        _, _, l_ave, _, _, _, _, _ = lbstats.print_function_statistics(
+        _, _, l_ave, _, _, _, _, l_imb = lbstats.print_function_statistics(
             curr_phase.get_ranks(),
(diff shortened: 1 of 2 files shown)


ppebay:
In passing @lifflander @nlslatt : is this `imbalance.txt` file (containing only the final value of the per-rank load imbalance) still needed? My recollection is that we wanted to use this for post-processing purposes, but the information is in the Exodus output anyway.

ppebay:
Thanks @nlslatt , I guess we don't remember it, it probably means we don't need it -- if @lifflander agrees I will also remove it as part of this PR.

cz4rs (approved):
Looks good to me.
Regarding the style, long-term solution would be to use some linters ([we already do that with Python code in `build-stats` repo](https://github.com/DARMA-tasking/build-stats/blob/master/.github/workflows/run-linter.yml)).

ppebay:
@cz4rs agreed -- in fact we had this somewhere in the "grand plan" for CI/CD but we never got to it I guess. Creating follow-on issue

PR #329, opened 2022-12-12
#327 Fix current impossibility to generate PNGs without EX2 and VTPs

### LINKED ISSUE #327: Fix current impossibility to generate PNGs without Ex2 and VTPs
As discussed with @lifflander it would allow for performance improvements when only PNGs are needed for subsequent analysis

ppebay:
Resolves #327


nlslatt (approved):
Looks good to me

PR #331, opened 2022-12-13
#330: create only necessary inter-object edges

### LINKED ISSUE #330: Create only used edges in inter-object communication graph mesh
This is following a discussion with @lifflander in order to improve performance (remove O(n-squared) edge insertion).

ppebay:
Resolves #330


ppebay:
Screen Shot 2022-12-13 at 1 01 12 PM

Rendered behavior unchanged but (much) fewer edges created:
```
[lbsMeshBasedVisualizer] Assembling object mesh with 9 points and 8 edges
```
instead of 9 * 8 / 2 = 36 edges

ppebay:
What I noticed while analyzing the above is that edges are repeated when there is traffic in both directions, cf 1--4 below (edges 1 and 4):

 
Screen Shot 2022-12-13 at 1 55 17 PM

ppebay:
Edges are now sum-reduced when there is traffic in both directions:


Screen Shot 2022-12-14 at 3 57 08 PM

ppebay:
Also working now for null edge meshes (as we no longer have the full, fully-connected edge diagram as before)

Screen Shot 2022-12-14 at 10 02 49 PM

lifflander (approved):

PR #341, opened 2023-01-26
#340: Implement clustering-based transfer strategy

### LINKED ISSUE #340: Implement clustering-based transfer strategy
This will be another concrete type of transfer strategy, and at first it will hard-code the shared memory block ID based clustering

ppebay:
Resolves #340


ppebay:
The current implementation of the clustering-based approach, that is also breaking down clusters when needed, is now capable of finding the global optimum (max load = 90), but at the expense of allowing for a **transient** overshoot on the max memory usage (while the final one remains under the actual contraint: < 8.0e+9; but must be allowed to go through 9.1e+9 during one intermediate-step transfer): ` max_memory_usage: 9.1e+9`

Screen Shot 2023-01-31 at 11 29 34 AM

Screen Shot 2023-01-31 at 11 29 41 AM

I think @lifflander this shows that your centralized grouping-based approach can be replicated in a distributed setting, but with some transient upper-bound on the bounded QOI constraint relaxation. What do you think?

ppebay:
Fixed bug in rank QOI color scheme that I just noticed; this is the correct color-coded visualization of the last iteration:
Screen Shot 2023-01-31 at 12 32 27 PM

ppebay:
And this is what the corresponding of max memory usage look like:

Initial (uniform):
Screen Shot 2023-01-31 at 12 38 58 PM


First iteration (transient overshoot on rank 1):
Screen Shot 2023-01-31 at 12 39 04 PM


Final (max memory bound back within prescribed bounds after iteration 2):
Screen Shot 2023-01-31 at 12 39 10 PM

@lifflander @nlslatt removing Draft marking so first review can be done. Thanks

ppebay:
Method now converging in a single step on reference test case:

Screen Shot 2023-02-13 at 7 24 24 PM
Screen Shot 2023-02-13 at 7 24 34 PM

Ready for review @lifflander @nlslatt thanks

ppebay:
Validating that this is also working optimally for the non-deterministic transfer (i.e. using empirical PMF-based target sampling):
Screen Shot 2023-02-14 at 10 24 56 AM
Screen Shot 2023-02-14 at 10 25 05 AM

ppebay:
@nlslatt @lifflander with swapping on, converging consistently towards near-optimum on challenging problems (fewer tasks):
Screen Shot 2023-02-19 at 11 44 28 AM

ppebay:
With `thousands_tasks`, initial imbalance:

Screen Shot 2023-02-19 at 1 49 51 PM
Screen Shot 2023-02-19 at 1 44 47 PM

Imbalances for 8 successive iterations with **partial information** (~4 known peers: fanout=2, 2 rounds):

Screen Shot 2023-02-19 at 1 54 16 PM

Screen Shot 2023-02-19 at 1 49 40 PM

ppebay:
Further improvements to the subsampling strategy (the "non-deterministic" option) for large cases:

Screen Shot 2023-02-21 at 7 56 24 PM


Screen Shot 2023-02-21 at 7 56 38 PM

ppebay:
@lifflander @nlslatt PR ready for review

ppebay:
Following discussion with @nlslatt , "hundreds" case, with 64 ranks, full information, 8 iterations:
Screen Shot 2023-02-22 at 7 52 16 PM
Screen Shot 2023-02-22 at 7 51 44 PM

After 8 **non-deterministic** iterations:
Screen Shot 2023-02-22 at 7 52 27 PM
Screen Shot 2023-02-22 at 7 51 53 PM

ppebay:
Results with the `thousands` example, on 256 ranks:

Screen Shot 2023-02-23 at 4 39 48 PM
Screen Shot 2023-02-23 at 4 39 25 PM

After only 4 iterations with rather limited information (k=2, f=4), we get the following results with the (newest) subsampling strategy:

Screen Shot 2023-02-23 at 4 41 32 PM
Screen Shot 2023-02-23 at 4 41 13 PM

lifflander (approved):
Looks great to me. I see that you are trying swaps first! I think that's a big difference between this and my algorithm. Perhaps that makes the most sense.

ppebay:
@tlamonthezie can you please look into what's causing the unit tests to fail? Thanks

tlamonthezie:
Hello @ppebay.
This is an issue with the `ubuntu-latest` docker image which is used to run the unit tests. The issue is happening on any push running the unit tests (not only this PR).
A possible solution to not have this error is to go to another version. I tested the `ubuntu-20.04` image and it works correctly.

tlamonthezie:
@ppebay fix it by going to .github/workflows/unit-tests.yml
then add the following command `sudo apt-mark hold grub-efi-amd64-bin && sudo apt-mark hold grub-efi-amd64-signed` to prevent the upgrade of the `grub-efi-amd64-bin` and the `grub-efi-amd64-signed` packages (l24)

In fact the upgrade is of these packages is the reason of failure.

image

Issue #344, opened 2023-02-23
Make object QOI keyword optional in configuration validator and test harness


ppebay:
This is because we now want to allow the user to **not** generate the object mesh when there are too many objects (which become invisible anyway). However if it's not passed in the configuration file, this makes the test harness fail.

ppebay:
Resolved with #346

PR #346, opened 2023-02-24
#343: Changes to visualizer for large object counts

### LINKED ISSUE #343: Changes to visualizer for large object counts
Add the following features:
* ability to adjust object scaling as extra parameter
* make it possible to deactivate entirely the object mesh creation (esp. for large object counts)
* allow for turning off of VOM exports

ppebay:
Resolves #343 #344


ppebay:
Case in point: "hundreds" case; `object_qoi` **not** specified in the configuration file (as it's now optional), only rank mesh is generated, with attendant much faster LBAF-Viz output (also note that a new `write_vt` parameter has also been made available, optional as well, and `True` by default for BW-compatibility with the host of existing configuration files; this also massively reduces output time when turned off for such large cases):

Screen Shot 2023-02-25 at 1 23 18 PM
Screen Shot 2023-02-25 at 1 23 31 PM

tlamonthezie (approved):
Unit tests check is failing but fixed in develop so it's fine

Issue #336, opened 2023-01-04
Make configuration validation for YAML files less painful to update


lifflander:
Right now, it requires a lot of changes in multiple files when a key is changed.

tlamonthezie:
@lifflander, @ppebay  

1. Could you please give me an example scenario of a change ?
This is the main point: when the configuration schema changes, which files do you change and then which ones need to be updated. Do you first the `ConfigurationValidator` ?

We can envisage different possibilities:
- What about some python script to analyze the changed schema and list the files that needs to update (`data/configuration_examples/*.yaml`, `src/lbaf/Applications/*.yaml`, `test/data/config/*.yaml`, and maybe classes to update...)
- We coul also tag each yaml configuration file with some version number that can help to list the files that needs to be updated (for yaml files). This current version number could be defined at the `ConfigurationValidator` level.
The discussion is open...

3. The following might be linked to this issue: in the `internalParameters` class (`LBAF_app.py`) there is some kind of duplication of the allowed keys. When defining a new key in `ConfigurationValidator` we need also to define it in `internalParameters` (for keys at root level). I think we can avoid that.

4. Some pylint errors appears in the 'internalParameters' class like the following I don(t know if it can help but it will be good
`src/lbaf/Applications/LBAF_app.py:179:43: E1101: Instance of 'internalParameters' has no 'n_ranks' member (no-member)`

ppebay:
Thanks for the detailed question @tlamonthezie 
Regarding 1.: try for instance to add one keyword parameter to `conf.yaml`, for instance `foo: bar`. You should see the error it's causing in the configuration validator.

Regarding 3. I agree, there is some redundancy there which I had noticed before and was curious about. It should be eliminated, that said I am not sure where it's best to leave it, either in `LBAF_app.py` or in the configuration validator. Thoughts?

4. Yes please feel free to clean up these issues when you encounter them.


Thanks!

tlamonthezie:
@ppebay thank you for your response.

He is a more advanced analysis:

## Scenarios
Considering that at first we edit `conf.yaml` to defie a change in the schema

### Scenario 1 : add new key##
1. You get a `SchemaError` when running the app. This is good because the configuration does not match the schema (and the schema defines the allowed structure).
2. Then you will update the schema in the `ConfigurationValidator` to setup the key specifications inside the schema - and if root key add to allowed config keys in `internalParameters`  - to resolve this error _
3.a The key is optional : it works immediately
[...]

PR #350, opened 2023-03-01
#336: Make configuration validation for YAML files less painful to update

tlamonthezie:


config/user-defined-memory-toy-problem.yaml
```diff
@@ -1,21 +1,20 @@
...
   parameters:
-    alpha: 1.
-    beta: 0.
-    gamma: 0.
+    alpha: 1.0
+    beta: 0.0
+    gamma: 0.0
     upper_bounds:
-      max_memory_usage: 8.0e+9
+      max_memory_usage: 8000000000.0
```
ppebay, line 15: Why this format?

tests/config/conf_wrong_from_data_phase_type.yml
```diff
@@ -1,18 +1,18 @@
 # Specify input
 from_data:
-  data_stem: "../data/synthetic_lb_data/data"
+  data_stem: ../data/synthetic_lb_data/data
   phase_ids:
-    - 0.0
+  - 0.0
```
ppebay, line 5: This should be indented, shouldn't they?

tests/config/conf_wrong_data_and_sampling.yml
```diff
@@ -38,18 +38,18 @@ LBAF_Viz:
...
 from_samplers:
   n_objects: 200
   n_mapped_ranks: 2
   communication_degree: 20
   load_sampler:
     name: lognormal
     parameters:
-      - 1.0
-      - 10.0
+    - 1.0
```
ppebay, line 49: Why did you remove the indentation here?

tests/config/conf_wrong_data_and_sampling.yml
```diff
@@ -23,13 +23,13 @@ algorithm:
     transfer_strategy: Recursive
     criterion: Tempered
     max_objects_per_transfer: 8
-    deterministic_transfer: True
+    deterministic_transfer: true
```
ppebay, line 26: Is the capitalized `True` non-standard and to be replaced by `true`?

src/lbaf/Utils/functions.py
```diff
@@ -0,0 +1,17 @@
+"""Utility functions"""
+import os
+from typing import Union
+
+def abspath_from(path: Union[str, None], relative_to_dir: str) -> Union[str, None]:
+    """Get the absolute path from the given path relative to specific directory.
```
ppebay, line 6: Add space here after triple double quotes

scripts/upgrade_configuration.py
```diff
@@ -0,0 +1,184 @@
...
+                else:
+                    node[key] = parsed_value
+            # go next node in child tree
+            if is_leaf:
+                break
+            else:
+                node=node[key]
+
+    def write_node(k, value, yaml_file):
+        """Wirtes a single node (key/value) in the given yaml file"""
```
ppebay, line 119: Typo (Wirtes/Writes); no s at end of very; missing space before W

scripts/upgrade_configuration.py
```diff
@@ -0,0 +1,184 @@
...
+logger.info('Script to bulk add or remove key to/from LBAF configuration files within the project')
+logger.info(
+    'NOTICE: Remember that the keys must be defined first at the schema level defined in ' \
+    'the ConfigurationValidator class'
+)
+
+def upgrade(file_path: Path) -> int:
+    """
+    Apply an upgrade to the given configuration file
+    """
```
ppebay, line 66: Docstring formatting not consistent with the rest

scripts/upgrade_configuration.py
```diff
@@ -0,0 +1,184 @@
+"""
```
ppebay, line 1: Inconsistent docstring formatting

ppebay (changes requested):
Agreed with general approach but multiple instances of the same remarks (therefore not repeated multiple times):
* docstring formatting not consistent
* YAML file indentations has been changed before hyphens: why? 
* typos

scripts/upgrade_configuration.py
tlamonthezie: Ok thanks it will be corrected

tlamonthezie:
@ppebay regarding docstrings, it seems to be inconsistent also in other files so my question is
Is there some existing detailed doctring syntax for this project (modules, classes, methods, parameters etc.) ?

config/user-defined-memory-toy-problem.yaml
tlamonthezie: The file has been rewritten by the yaml dumper like all the configuration files. 
After some research it is due to Python standard formatting. In fact, Python will display values as scientific notation when they are bigger than 1.0e+15 or smaller than 1.e-4 this is internal to python formatting. And pyyaml output format is just the standard format. If you have a value of 8.0e+16 for example it will be formatted using the scientific notation.

tests/config/conf_wrong_data_and_sampling.yml
tlamonthezie: pyyaml does it but as written previously we will fix it using a workaround

tests/config/conf_wrong_data_and_sampling.yml
tlamonthezie: This is the standard way used by pyyaml to dump bool values

scripts/upgrade_configuration.py
tlamonthezie: Thanks @ppebay it will be corrected

src/lbaf/Utils/functions.py
tlamonthezie: @ppebay are you sure ? in doctring python documentations I often see there is no space like there: https://peps.python.org/pep-0257/

scripts/upgrade_configuration.py
ppebay: I don't think we have any specific ones necessarily, they just need to be consistent throughout. Pick whichever one you prefer and can be easily made consistent

ppebay:
@tlamonthezie Thanks for the detailed summary and proposal.

I would vote for rST as this would also be consistent with other projects.

tlamonthezie:
> @tlamonthezie Thanks for the detailed summary and proposal.
> 
> I would vote for rST as this would also be consistent with other projects.

@ppebay I have just commited some refactored files including also some consistency with docstrings with some respect of PEP conventions.

Some notes: After some investigations it seems that m.css is not parsing correctly doctrings for method parameters but it is fine with modules, classes and methods summary. As I know the m.css process has not run for some time and it seems that using sphinx might be better than m.css in the future (used in many projects, integrated with many IDEs). It might represent a future issue if we decide to move to sphinx.

tlamonthezie:
> Thanks @tlamonthezie can you please re-request a review from me?
> 
> As there are multiple changes it's best that I re-review from scratch

May be @ppebay . Did you see my remark regarding scientific notations ?

ppebay (approved):

.gitignore
```diff
@@ -16,3 +16,5 @@ src/Applications/tcl
 /output
 /build
 src/lbaf/imported
+/m.css
+/docs/output
```
ppebay, line 20: No new line at end of file

ppebay (approved):
Looking good now

.gitignore
tlamonthezie: Corrected

ppebay:
@tlamonthezie you will need to rebase to resolve the conflicts, then you can approve
Thanks

tlamonthezie:
> @tlamonthezie you will need to rebase to resolve the conflicts, then you can approve Thanks

@ppebay I resolved conflicts it seesm to be ok now

ppebay:
Thanks feel free to merge and delete branch @tlamonthezie

ppebay:
@tlamonthezie can you please elaborate a bit regarding the usage of the `-p` flag (for patterns)? Thanks.

PR #369, opened 2023-03-29
#366: Make threshold between categorical and continuous color scales a user parameter

ppebay:
Resolve #366

### DIFF
--- a/config/acceptance_tests/conf.yaml
+++ b/config/acceptance_tests/conf.yaml
@@ -32,5 +32,3 @@ terminal_background: light
 output_dir: ../../output
 output_file_stem: output_file
 n_ranks: 4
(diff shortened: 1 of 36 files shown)


ppebay:
Toy problem with `force_continuous_object_qoi` set to `true`:


Screen Shot 2023-03-30 at 1 46 17 PM


Absent `force_continuous_object_qoi`, **or** set to `false`, the object QOI is considered categorical as a reason of its support having less than 20 distinct values:
Screen Shot 2023-03-30 at 1 46 34 PM

nlslatt (approved):
In cases where the qoi is categorical, can you make the categories integers (i.e., not print the decimal part) so that they fit better?

PR #374, opened 2023-04-15
#367: make outputJSON files compatible with vt offline LB

### LINKED ISSUE #367: Make output JSON files compatible with vt's OfflineLB
The JSON files output by LBAF will be consumed by vt's `LBDataRestartReader` and replayed (in terms of object locations) by vt's `OfflineLB`. There are certain conditions that need to be met by LBAF's output files in order for the files to be acted upon as expected. I'll start with two common use cases before explaining more generally.

**An input file containing only phase 0:** LBAF should write the input phase 0 data verbatim into the output file, still labeled as phase 0. The phase 1 data written to the output file should contain loads from phase 0 but use the post-LB object locations.

**An input file containing phases 0 and 1, where LBAF only load balances on phase 0:** LBAF should write the input phase 0 data verbatim into the output file, still labeled as phase 0. The phase 1 data written to the output file should contain loads from the phase 1 inputs but use the post-LB object locations.

**For cases where there are more phases of data in the input file, LBAF balances more than one phase, or LBAF starts load balancing not on phase 0:**
- All phases before and including the first one LBAF attempts to load balance should be copied verbatim from the input file into the output file.
- The load balancer input for any given phase `J` should be the object locations and loads that were already written to the output file for phase `J` (if the load balancer has already run on any phase, this will likely not match the object locations from the input file).
- The output of the load balancer that was run on phase `J` will be written to the output file as phase `J+1`. If phase `J+1` existed in the input file, use the loads specified (but not object locations) in the input file for writing phase `J+1`; otherwise, re-use the loads from the phase `J` input.
- If the load balancer was not run on some phase `K`, the output for phase `K+1` will have the same object locations as phase `K` but use the loads from the phase `K+1` input.

Does that all sound correct to you @lifflander ?

ppebay:
Resolves #367

### DIFF
--- a/src/lbaf/Applications/LBAF_app.py
+++ b/src/lbaf/Applications/LBAF_app.py
@@ -398,8 +398,8 @@ class LBAFApp:
             vt_writer = VTDataWriter(
                 curr_phase,
                 self.logger,
-                self.params.output_file_stem,
-                output_dir=self.params.output_dir)
+                self.params.output_dir,
+                self.params.output_file_stem)
             vt_writer.write()
 
         # Generate meshes and multimedia when requested
--- a/src/lbaf/Applications/LBAF_app.py
+++ b/src/lbaf/Applications/LBAF_app.py
@@ -25,8 +25,8 @@ from lbaf.Utils.colors import green
 from lbaf.Utils.functions import abspath_from
 # pylint: enable=C0413
 
-def get_config_file() -> str:
-    """Parses command line argument and returns config file path."""
+def get_config_file_path() -> str:
+    """Parse command line argument and return config file path."""
     parser = argparse.ArgumentParser()
     parser.add_argument("--config", help="Path to the config file.", default="conf.yaml")
     args = parser.parse_args()
@@ -74,7 +74,7 @@ def check_and_get_schema_validator():
     overwrite_validator = True
     config_file = None
     if __name__ == "__main__":
-        config_file = get_config_file()
+        config_file = get_config_file_path()
         with open(config_file, "rt", encoding="utf-8") as config:
             conf = yaml.safe_load(config)
         overwrite_validator = conf.get("overwrite_validator", True)
@@ -117,7 +117,7 @@ class InternalParameters:
     rank_qoi: Union[str,None]
     object_qoi: Union[str,None]
     grid_size: Union[list,None]
-    write_vt: bool
+    vt_writer: Union[VTDataWriter,None]
 
     # from_samplers options
     n_objects: int
@@ -228,7 +228,6 @@ class InternalParameters:
                 self.phase_ids = list(range(range_list[0], range_list[1] + 1))
             else:
                 self.phase_ids = config.get("from_data").get("phase_ids")
-        self.write_vt = config.get("write_vt", True)
 
         # Parse sampling parameters if present
         if config.get("from_samplers") is not None:
@@ -241,6 +240,9 @@ class InternalParameters:
         # Set output directory, local by default
         self.output_dir = abspath_from(config.get("output_dir", '.'), config_dir)
 
+        # Determine whether VT JSON file outputs are requested
+        self.write_vt = config.get("write_vt", True)
+
     def check_parameters(self):
         """Checks after initialization."""
         # Checking if output dir exists, if not, creating one
@@ -251,13 +253,20 @@ class InternalParameters:
 class LBAFApp:
     """LBAF application class"""
     def __init__(self):
-        self.config_file = get_config_file()
+        # Retrieve configuration file path
+        self.config_file = get_config_file_path()
 
         # Instantiate parameters
         self.params = InternalParameters(config_file=self.config_file)
 
         # Assign logger to variable
         self.logger = self.params.logger
+
+        # Create VT writer except when explicitly turned off
+        self.vt_writer = VTDataWriter(
+            self.logger,
+            self.params.output_dir,
+            self.params.output_file_stem) if self.params.write_vt else None
 
     def main(self):
         """LBAFApp entrypoint to run"""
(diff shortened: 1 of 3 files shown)


ppebay:
Validating that the `PhaseStepper` is indeed working with the refactored logic:

https://user-images.githubusercontent.com/17439260/233868172-3b95c30f-c44c-4fc1-917d-a96478e23231.mp4

ppebay:
Verifying consistency of phase vs. iteration labels in the context of the new multi-phase logic:


https://user-images.githubusercontent.com/17439260/233868445-2599d8c2-b2c9-457c-b6a6-c8c333ab9d09.mp4

ppebay:
* Validating that this is now also working when there are shared blocks (following bug fix above); first the balancing process (4 iterations of `InformAndTransfer` algorithm with `Clustering` strategy:

https://user-images.githubusercontent.com/17439260/233872579-911a12db-4f50-4471-8578-dc4a162cd1ef.mp4


* Subsequently we can see the rendering of the 2 phases of the JSON files produced by the above, when visualized with the `PhaseStepper`:

test_file_00
test_file_01

ppebay:
This PR is now ready for review; it has actually turned into somewhat of a refactoring, due to the fact that the initial design only considered in-place rebalancing, for the sake of execution speed and memory efficiency.

In order to achieve the desired goals of this issue, the underlying model for the `lbsPhase` had to be entirely revisited, and as a result the `lbsRuntime` and algorithm classes as well. There was therefore much more to the story than just changing the VT JSON writer, and for this reason reviewers will have much more work than anticipated to review this PR ;) !

src/lbaf/IO/lbsVTDataWriter.py
```diff
@@ -17,35 +16,34 @@ class VTDataWriter:
...
-    def write(self):
-        """ Write one JSON file per rank."""
-        sys.setrecursionlimit(25000)
-        with Pool(context=get_context("fork")) as pool:
-            results = pool.imap_unordered(self.json_writer, self.__phase.get_ranks())
-            for file_name in results:
-                self.__logger.info(f"Saved {file_name}")
-
-    def __create_object_entries(self, rank_id, objects):
+    def __create_tasks(self, rank_id, objects):
```
nlslatt, line 46: This still outputs only a subset of the potential task attributes that may have been included in the input file. If something appears in the input file, I think we should retain it even if LBAF doesn't know what it means. What is the plan for moving in that direction?

src/lbaf/IO/lbsVTDataWriter.py
```diff
@@ -58,32 +56,66 @@ def __create_object_entries(self, rank_id, objects):
...
-            r_id, rank.get_sentinel_objects())
-        phase_data["tasks"] = tasks
-
-        # Write file and return its name
-        json_str = json.dumps(output, separators=(',', ':'))
-        compressed_str = brotli.compress(
-            string=json_str.encode("utf-8"), mode=brotli.MODE_TEXT)
-        with open(file_name, "wb") as compr_json_file:
-            compr_json_file.write(compressed_str)
+            "phases": []}
```
nlslatt, line 73: Again, not everything that appears in an input file is being output here. There can be more content to the metadata than is shown.

src/lbaf/IO/lbsVTDataReader.py
```diff
@@ -91,79 +109,158 @@ def _load_vt_file(self, rank_id: int):
...
+        rank_blocks, task_user_defined = {}, {}
+
+        # Iterate over tasks
+        for task in phase.get("tasks", []):
+            # Retrieve required values
+            task_entity = task.get("entity")
+            task_id = task_entity.get("id")
+            task_load = task.get("time")
+            task_user_defined = task.get("user_defined", {})
+            subphases = task.get("subphases")
```
nlslatt, line 187: I think anything that wasn't explicitly retrieved above should still be retained for output purposes. Maybe add another argument to the object constructor that contains unrecognized parts of the dict?

nlslatt (commented):
Will anything in this PR cause a failure for cases where objects can be created or deleted on any phase?

This looks like great progress, but it still does not seem to read in *everything* in the json that pertains to task attributes or metadata and retain it for output.

src/lbaf/IO/lbsVTDataWriter.py
ppebay: Good question! I think we need to discuss what we want to do with it, whether we want it to be fully-fledged or not (given the fact that the vt C++ version will supersede it). Can we put this on the agenda for the upcoming LB discussion?

src/lbaf/IO/lbsVTDataReader.py
ppebay: Following our discussion of 05/03 I believe this is is resolved

src/lbaf/IO/lbsVTDataWriter.py
ppebay: Following our discussion of 05/03 I believe this is is resolved

src/lbaf/IO/lbsVTDataWriter.py
ppebay: Following our discussion of 05/03 I believe this is is resolved

ppebay:
Hello dear reviewers @lifflander @nlslatt @pierrepebay this is a reminder :)

src/lbaf/Applications/LBAF_app.py
pierrepebay, line 1: Is `self.write_vt` still used ? I see it has been removed from the internal parameters and the only place I see this is line 214 in `self.write_vt = config.get("write_vt", True)`

ppebay: Good point. It is now superseded by the `write_JSON ` key in the configuration file; and the above in line 214 was indeed a leftover (now removed).

pierrepebay:
In `rank_object_enumerator.py`, the `LoadReader` is not initialized with correct arguments and calls the `read` function which doesn't exist anymore:
https://github.com/DARMA-tasking/LB-analysis-framework/pull/374/files#file-src-lbaf-applications-rank_object_enumerator-py-L30
https://github.com/DARMA-tasking/LB-analysis-framework/pull/374/files#file-src-lbaf-applications-rank_object_enumerator-py-L34
 It appears only the `compute_min_max_arrangements_work` function from this file is used elsewhere, so I'm not sure if this is very relevant.

PR #378, opened 2023-04-26
#377: GitHub action passes even when python error is raised in tox run

### LINKED ISSUE #377: GitHub action passes even when Python error is raised in tox run
Relative import path Python error fixed in #373 was not causing the GitHub actions to be marked as failing: https://github.com/DARMA-tasking/LB-analysis-framework/actions/runs/4569049204/jobs/8064831145#step:7:1627, if other errors appear will the tox run be marked as passing as well ?

tlamonthezie:
Fixes #377 

Error was occuring because of relative imports in unit tests when calling during coverage step (it did not impact the unittests)

The addition of an argument in tox.ini finally fixed the error. See also more at https://github.com/DARMA-tasking/LB-analysis-framework/issues/377#issuecomment-1523330790


ppebay (approved):
Working for me

PR #389, opened 2023-05-26
#355: Test the VT writer

### LINKED ISSUE #355: Test the VT writer
* We need to make sure that VT outputs are identical to the inputs when no algorithm is executed ("null test")
* We need to validate VT outputs against the JSON schema validator 

**N.B.:**  these are merely necessary conditions, certainly not sufficient, but nevertheless important

tlamonthezie:
Fixes #355 

This PR include the addition of a new unit test to the project tests to check that the input data and the output data contain the same mandatory data when using a StepperTest algorithm.

Note: To be able to compare input and output data it is first cleaned and reordered :
1. optional keys are auto-detected (with the help of the schema definition) 
2. then optional keys are removed from input and output data
3. data is sorted
It prevents this test to fail in case keys or nodes are not ordered the same way which is currently the case

Additionally it solves a problem with the VTDataWriter which was not correctly writing the migratable values, brings some pylint additional configuration + sort of .gitignore lines to help to prevent line duplication


ppebay (approved):
Implements desired test behavior

Issue #154, opened 2022-01-31
Bug: logger printing white characters in the message body


ppebay:
White on white is not very easy to read:
<img width="407" alt="Screen Shot 2022-02-01 at 12 54 04 AM" src="https://user-images.githubusercontent.com/17439260/151891954-cbfeaa71-6573-4533-844b-1913985fe586.png">

PR #391, opened 2023-05-30
#347: Fix terminal background in configuration file

### LINKED ISSUE #347: Fix terminal background option in configuration file
The `terminal_background` option is not portable. We need to find a better solution.

tlamonthezie:
Fixes #347 

- This pull request remove the terminal_background option from the configuration and from the existing configuration files.
 Note: we might need to check again for #154 . If the problem of #154 occurs again it should be resolved by client user by changing the terminal settings. That is because a logged message is written to the console after a "clear color" so it should use the default user terminal color. If color is bad then change the client user terminal text color
- terminal_background option has also been removed from rst docs and from the schema defined in the ConfigurationValidator class
- Additionnaly it improves the configuration_upgrader.py script to keep comments at the top of the file if defined
-  It also adds the display of an error message and exits lbaf if running with Python 3.9 in case save_meshes is True in configuration file from the Visualizer. This is to explicitly say that this error is known.
- LBAF_app refactoring for more readability : load arguments inside a dedicated method
- Fix some configuration parameters not classified in sections in configuration files when using configuration upgrader

### DIFF
--- a/src/lbaf/IO/lbsConfigurationUpgrader.py
+++ b/src/lbaf/IO/lbsConfigurationUpgrader.py
@@ -23,6 +23,15 @@ from .. import PROJECT_PATH
 #     return dumper.represent_scalar(u'tag:yaml.org,2002:float', text)
 # yaml.add_representer(float, float_representer)
 
+def bool_representer(dumper, value):
+    """Overrides default yaml representation of boolean values for the ConfigurationDumper"""
+
+    if value:
+        text = 'True'
+    else:
+        text = 'False'
+    return dumper.represent_scalar('tag:yaml.org,2002:bool', text)
+
 class UpgradeAction(Enum):
     """Upgrade action"""
 
@@ -44,6 +53,7 @@ class ConfigurationUpgrader:
 
     def __init__(self, logger: Logger):
         self.__dumper = ConfigurationDumper
+        self.__dumper.add_representer(bool, bool_representer)
         self.__logger = logger
         self.__sections = cast(dict, ConfigurationValidator.allowed_keys(group=True))
 
@@ -67,7 +77,7 @@ class ConfigurationUpgrader:
             if yaml_node.endswith("\n" + indent_str):
                 yaml_node = yaml_node[:-(indent_size + 1)]
         else:
-            yaml_node = " " + yaml.representer.SafeRepresenter().represent_data(value).value
+            yaml_node = " " + yaml.representer.BaseRepresenter().represent_data(value).value
         yaml_file.write(yaml_node)
         yaml_file.write("\n")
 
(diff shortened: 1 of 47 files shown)


src/lbaf/IO/lbsConfigurationUpgrader.py
```diff
@@ -23,6 +23,15 @@
 #     return dumper.represent_scalar(u'tag:yaml.org,2002:float', text)
 # yaml.add_representer(float, float_representer)
 
+def bool_representer(dumper, value):
+    """Overrides default yaml representation of boolean values for the ConfigurationDumper"""
+
+    if value:
+        text = 'True'
```
ppebay, line 30: Please use double quotes for strings

src/lbaf/IO/lbsConfigurationUpgrader.py
```diff
@@ -23,6 +23,15 @@
...
+    else:
+        text = 'False'
```
ppebay, line 32: Same as above

src/lbaf/IO/lbsConfigurationUpgrader.py
```diff
@@ -129,6 +159,9 @@ def upgrade(
                 node=node[key]
 
         with open(file_path, "w", encoding="utf-8") as yaml_file:
```
ppebay, line 161: Simple quotes for single characters. Please enforce throughout entire code base

src/lbaf/Utils/logging.py
```diff
@@ -13,87 +14,116 @@
...
 
 class CustomFormatter(Formatter):
-    """Custom formatter class defining a format by logging level """
-    formatters: Dict[int,Formatter] = {}
-    def __init__(self, frmttr):
-        super(CustomFormatter, self).__init__()
-        for level, fmt in frmttr.items():
-            self.formatters[level] = Formatter(fmt)
+    """Formatter able to write colored logs
+    Colors are used to colorize log meta information such as
```
ppebay, line 29: add colon before starting itemization

src/lbaf/Utils/logging.py
```diff
@@ -13,87 +14,116 @@
...
-    logger = d.get(name, None)
-    if logger is not None:
-        return logger
+    # logger = logging.Logger.manager.loggerDict.get(name)
+    # if logger is not None:
+    #     return logger
+    if name in loggers:
+        return loggers[name]
+
+    print("Init logger " + name + "...")
```
ppebay, line 107: Write "Initializing" instead of "Init" and remove triple dots

tests/config/conf_wrong_from_samplers_load_sampler_003.yml
```diff
@@ -1,4 +1,18 @@
 # Specify input
+from_samplers:
+  n_objects: 200
+  n_mapped_ranks: 2
+  communication_degree: 20
+  load_sampler:
+    name: lognormal
+    parameters:
+      - 1.0
+      - '10.0'
```
ppebay, line 10: double quotes

ppebay (changes requested):
Changes requested

src/lbaf/Utils/logging.py
tlamonthezie: finally print removed

tlamonthezie:
> Changes requested

Thank tou @ppebay should be now ok

ppebay (approved):

PR #405, opened 2023-06-14
#388: print list of implemented qois

### LINKED ISSUE #388: Print list of implemented QOIs
If it is does not yet exist, add a command-line option to LBAF to print a list of implemented QOI. When a yaml file contains an invalid QOI, print that list as part of the error message. The current error handling approach (the Python script erroring out because a function has not been defined) is not at all user-friendly.

cwschilly:
Fixes #388

### DIFF
--- a/src/lbaf/Applications/LBAF_app.py
+++ b/src/lbaf/Applications/LBAF_app.py
@@ -270,11 +270,6 @@ class Application:
 
         return path
 
-    def __get_verbosity_level(self) -> int:
-        """Find the verbosity level from the '-verbosity' command line argument. When a yaml file contains an invalid QOI, print that list as part of the error message."""
-        verbosity = self.__args.verbose
-        
-
     def run(self):
         """Run the LBAF application."""
         # Parse command line arguments
@@ -299,9 +294,6 @@ class Application:
         JSON_data_files_validator_loader.load(cfg.get("overwrite_validator", True))
         if not JSON_data_files_validator_loader.is_loaded():
             raise RuntimeError("The JSON data files validator must be loaded to run the application")
-
-        self.__logger.warning("something")
-        self.__get_verbosity_level()
 
         # Initialize random number generator
         lbstats.initialize()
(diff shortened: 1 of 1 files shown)


cwschilly:
Here is the (end of the) output with `lbaf -v 2`
image

cwschilly:
Here is the output for `lbaf -v 1`
image

src/lbaf/Applications/LBAF_app.py
```diff
@@ -224,6 +224,10 @@ def __parse_args(self) -> dict:
                 "directory or the config directory",
             default=None
         )
+        parser.add_argument("-v", "--verbose",
+            help="Verbosity level. If 1, print the rank QOI. If 2, print the rank QOI and the object QOI.",
```
ppebay, line 228: I like this

src/lbaf/Applications/LBAF_app.py
```diff
@@ -463,9 +467,27 @@ def run(self):
...
         self.__logger.info("Process completed without errors")
 
+    def __print_QOI(self) -> int:
+        """Print list of implemented QOI based on the '-verbosity' command line argument."""
+        verbosity = int(self.__args.verbose)
+
+        # Print QOI based on verbosity level
+        if verbosity == 1:
+            self.__logger.info(f"List of implemented QOI \n \
+            Rank QOI:   {self.__parameters.rank_qoi}"
```
ppebay, line 483: We want **all possible ones**, not only the ones in the parameters

ppebay (changes requested):
I like the -v 1 or 2 to determine whether or not we want to drill down to the object level.

However we need to show all possible QOIs (basically all the quantities for which a `get_...()` method is implemented in `lbsRank` and `lbsObject`. So this is dynamic as the code must be analyzed to find out.

cwschilly:
I have altered the code so it outputs all possible QOI. Here is the new output with `lbaf -v 2`
image

ppebay:
@cwschilly changes looking good!

Minor formatting suggestions:
1- use `Rank:` and `Object:` (instead of the capitalized versions thereof)
2- using a second level of indentation for the actual QOI names under `Rank:` and `Object`

ppebay (approved):
All good now!

PR #406, opened 2023-06-16
#359: Replace list return types with sets wherever possible

### LINKED ISSUE #359: Replace list return types with sets wherever possible
This is to improve performance: it is faster to iterate over set than over lists and even faster to check element ownership.

There are cases where we want to perform indexing and/or modification of the returned collection and a result in those cases lists should be conserved.

cwschilly:
Fixes #359

### DIFF
--- a/src/lbaf/IO/lbsConfigurationUpgrader.py
+++ b/src/lbaf/IO/lbsConfigurationUpgrader.py
@@ -140,8 +140,8 @@ class ConfigurationUpgrader:
 
             yaml_file.seek(0)
             yaml_content = yaml_file.read()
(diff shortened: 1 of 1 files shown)


cwschilly:
I have not found any situations in lbaf where converting to a set would make the code perform better. I thought I had found one (switching `conf` to a set in `lbsConfigurationUpgrader.py`) but it did not end up improving the code so I undid the commit. It may be suitable to close issue #359  with no files changed

ppebay:
Thank you @cwschilly for your investigations. 

Closing PR (and issue) accordingly

PR #407, opened 2023-06-16
#370: make categorical labels integers

### LINKED ISSUE #370: Make categorical labels integer when numeric
In cases where the QOI is categorical, make categories integers (i.e., not print the decimal part) so that they fit better

As suggested by @nlslatt

cwschilly:
Fixes #370

### DIFF
--- a/src/lbaf/IO/lbsVisualizer.py
+++ b/src/lbaf/IO/lbsVisualizer.py
@@ -453,12 +453,13 @@ class Visualizer:
         scalar_bar_actor.SetWidth(0.42)
         scalar_bar_actor.SetBarRatio(0.3)
         scalar_bar_actor.DrawTickLabelsOn()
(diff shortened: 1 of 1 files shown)


cwschilly:
This fix appears to work, but I'm not sure how to test with data that would yield decimal labels. @nlslatt , if you have a sample config file that produced the problem in #370 , please include it here so that I can test out this solution.

ppebay:
Any update @cwschilly on testing this PR?

cwschilly:
@ppebay No update yet; if there is a sample config file that produces the error, that would be great to be able to test on

ppebay:
Thanks @cwschilly I added @nlslatt as reviewer

cwschilly:
This will print labels like "1.0" and "2.0" as "1" and "2", but any other decimals (like "1.5") are still printed as such.

ppebay (approved):
Perfect

PR #408, opened 2023-06-19
#379: Non-blocking follow-on style and consistency post-PR358

### LINKED ISSUE #379: Non-blocking follow-on style and consistency post-PR358
This is a follow-on to PR #357 as some style/consistency issues were noted (but were not blocking for said PR):
* inconsistency in naming of utility classes (in `Utils`): they should start with `lbs`, with the exception of:
  *  stand-alone scripts (no particular style required for these)
  * imported scripts that come from elsewhere
* inconsistent use of final full stop before triple quotes in docstrings
* still some occurrences of exclamation marks in error/warning outputs

All of the above should be addressed as part of this issue

Reference : https://peps.python.org/pep-0008/#package-and-module-names

tlamonthezie:
Fixes #379

- Rename scripts with lbs prefix
- Refactor errors: add new TerseError class in lbsException.py (old exception_handler.py) and fix exception handler bad calls in code
- Fix pylint errors and warnings
- fix double quotes


src/lbaf/Execution/lbsAlgorithmBase.py
```diff
@@ -29,35 +27,24 @@ def __init__(self, work_model: WorkModelBase, parameters: dict, logger: Logger,
         """
         # Assert that a logger instance was passed
         if not isinstance(logger, Logger):
-            get_logger().error(
-                f"Incorrect type {type(logger)} passed instead of Logger instance")
-            sys.excepthook = exc_handler
-            raise SystemExit(1)
+            raise TerseError(f"Incorrect type {type(logger)} passed instead of Logger instance")
```
ppebay, line 30: What is the purpose of this `TerseError()`? Reduce code replication?

src/lbaf/Utils/lbsArgumentParser.py
```diff
@@ -0,0 +1,177 @@
...
+        """Asks a question"""
+        msg = green(question)
+        if default is not None:
+            msg += " [" + yellow(default) + ']'
+        else:
+            msg += " [" + blue(None) + ']'
+
+        raw_response = None
+
+        while raw_response is None:
```
ppebay, line 69: Would be nice to document this routine more

src/lbaf/Utils/lbsArgumentParser.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def set_args(self, args: dict, namespace=None):
+        """This method init updates a namespace with the default values and with the given arguments.
+
+        :param args: arguments dictionary to set specific argument values
+        """
+        if namespace is None:
+            namespace = argparse.Namespace()
+
+        # add any action defaults that aren't present
```
ppebay, line 119: Capitalize first letter of comments (everywhere)

src/lbaf/Utils/lbsPath.py
```diff
@@ -10,7 +10,7 @@ def abspath(path: str, relative_to: Optional[str] = None) -> Optional[str]:
 
     :param path: the input path
     :param relative_to: the base path, defaults to None (None = the current working directory)
-    :return: an absolute path
+    :returns: an absolute path
```
ppebay, line 13: not a good name for a path

ppebay (changes requested):

src/lbaf/Execution/lbsAlgorithmBase.py
tlamonthezie: Hello @ppebay this is due to refactoring.
In the code this was not consistent. Sometimes we raised errors and each time we do that after setting the exception handler. For a developer this is not intuitive. We should set the handler only once for an application and raise errors and then use the handler to process it if needed. Thats why many logger errors followed by system exit commands have been replaced by errors raising. Because it seems more logical for an error to be fired as a Python exception si it can if needed be intercepted. I think it is a better practice.

TerseError means no stack trace Just a simple error message to display. This is Just a way to know in the handler what we want to display. An alternative could be to detect only error raised by LBAF to be displayed with only the message then we could remove the TerseError class.

src/lbaf/Utils/lbsPath.py
tlamonthezie: @ppebay what do you suggest ?

ppebay: `absolute_path`?

src/lbaf/Execution/lbsAlgorithmBase.py
ppebay: I see your point about the codebase not being consistent with exception throwing and error-ing out.

That being said, the intent at least was the following:
* any condition resulting in an ERROR (and logged as such) should cause the code to terminate immediately; the rationale being that if a type-I error (i.e. a "known error" that can tested as such) occurs, this means that any subsequent result produced by the code could be incorrect; as a result the code should not be allowed to pursue further;
* in contrast, a condition merely resulting in a WARNING (and logged as such) could be allowed to proceed after having communicated the message as you described in the second paragraph above.

So, my point is that errors should always result in code termination. What you call a `TerseError` more resembles a warning in my definition.

src/lbaf/Utils/lbsPath.py
tlamonthezie: @ppebay Not sure it is logic to rename. In fact this function accepts both relative or absolute path but returns an an absolute path

ppebay:
Alright then. Thanks. On 2023. Jun 20., at 11:43, tlamonthezie ***@***.***> wrote:﻿
@tlamonthezie commented on this pull request.



In src/lbaf/Utils/lbsPath.py:
> @@ -10,7 +10,7 @@ def abspath(path: str, relative_to: Optional[str] = None) -> Optional[str]:
 
     :param path: the input path
     :param relative_to: the base path, defaults to None (None = the current working directory)
-    :return: an absolute path
+    :returns: an absolute path

@ppebay Not sure it is logic to rename. In fact this function accepts both relative or absolute path but returns an an absolute path

src/lbaf/Execution/lbsAlgorithmBase.py
tlamonthezie: TerseError removed and consistency added :
- logger error followed by SystemExit  everywhere we want to terminate with an ERROR.
- exception handler removed

ppebay (approved):

PR #409, opened 2023-06-19
#352: Create baseline/global YAML configuration file

### LINKED ISSUE #352: Create baseline/global YAML configuration file

tlamonthezie:
Fixes #352 

- Adds support of global configuration file if present at <project_path>/config/global.yaml or <cwd>/global.yaml 
- Add basic global.yaml file with output_dir default value of "../output". 

I think it will be important later to define what goes in global.yaml and what should be removed for some configurations files on the repo (like the "output_dir" key)

### DIFF
--- a/config/global.yaml
+++ b/config/global.yaml
@@ -1,0 +1,3 @@
+# global parameters uto use in case it is not defined in a configuration file
+
+output_dir: ../output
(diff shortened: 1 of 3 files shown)


ppebay (approved):
All good

PR #420, opened 2023-07-26
#418: Failure of test_vt_data_extractor_014 in CI

### LINKED ISSUE #418: Failure of test_vt_data_extractor_014 in CI
The `test_vt_data_extractor_014` test is failing on CI (apparently not on local machine WSL/Ubuntu)
This test has 2 assertions. Determine which assertion is failing and why and fix it.

tlamonthezie:
Fixes #418 

The test related to the vt_data_extractor was in fact failing due to a missing assertion possible value.
Now it works correctly.
Fix also some pylint warnings


ppebay (approved):
Approving the PR as the failure is fixed, but in the future it would be good to be consistent in the use of quotes/double quotes with the rest of LBAF

PR #425, opened 2023-08-29
#376: Enable multiple versions of lbaf to run in the same python environment

### LINKED ISSUE #376: Rework import structure in LBAF_app and other files
1. Reorganize Import statements and PYTHONPATH

In many files some module import are done by code using the `sys.path.append` functions but it will be cleaner if we use the PYTHONPATH environment variable like this:

`export PYTHONPATH="${PYTHONPATH}:/path/to/project"`

Currently many files contains a code like the following and with the environment variable it won't be needed

https://github.com/DARMA-tasking/LB-analysis-framework/blob/3720ef281e698b31e3bee339b2e94ad83cd59a34/src/lbaf/Applications/LBAF_app.py#L11-L16

After that import statements will be correctly grouped at the top of python modules it will be cleaner and more compliant with pylint 

2. Rework setup configuration to make setup work for Python 3.9 (currently setup requires Python 3.8)

tlamonthezie:
Fixes #376 

This PR 
- Restore the way to run LBAF not as a package. When package is not installed then the lbaf module will be resolved relative to the current project directory automatically.
- For users who want to run development mode using lbaf as a package they will have to run `pip install -e .` since it has been removed from requirements.txt
- README.md file has also been re-worked to explain how to use lbaf as a package & adds information about using virtual environments
- Tests are again run against the lbaf source directory (no more the package installation because it may not be installed as a package now)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,8 +10,7 @@ __pycache__
 /build
 /dist
 /m.css
-/venv
-/venv39
+/venv*
 output
 /src/Applications/Include
 /src/Applications/Lib
--- a/README.md
+++ b/README.md
@@ -28,11 +28,24 @@ Make sure you have all required Python packages installed with:
 pip install -r requirements.txt
 ```
 
+By default requirements installs also the lbaf package in editable mode as you can see by running `pip list`
+
 Requirements are divided into `LBAF dependencies` and `LBAF testing`.
 
 `LBAF dependencies` are needed in order to LBAF to work.
 
 `LBAF testing` are needed for testing purposes.
+
+## Support of multiple versions under the same python environment (at the same time)
+
+There is a particular case where you may need to manage multiple local lbaf installs on your machine at the same time.
+Since pip only supports 1 package version install, you will have to remove the pip dependency to lbaf itself by running:
+```shell
+pip uninstall lbaf
+```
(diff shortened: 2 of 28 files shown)


nlslatt:
This did not fix my problem. I tried uninstalling LBAF, checking out your branch, and running LBAF there, but got the same error as before:

```
  File "/Users/nlslatt/work/LB-analysis-framework/src/lbaf/Applications/LBAF_app.py", line 13, in <module>
    import lbaf.IO.lbsStatistics as lbstats
ModuleNotFoundError: No module named 'lbaf'
```

tlamonthezie:
> This did not fix my problem. I tried uninstalling LBAF, checking out your branch, and running LBAF there, but got the same error as before:
> 
> ```
> Traceback (most recent call last):
>   File "/Users/nlslatt/work/LB-analysis-framework/src/lbaf/Applications/LBAF_app.py", line 13, in <module>
>     import lbaf.IO.lbsStatistics as lbstats
> ModuleNotFoundError: No module named 'lbaf'
> ```

@nlslatt Sorry there was an remaining invalid path in LBAF_app.py. This should work now. Could you please retry ?

nlslatt:
@tlamonthezie @ppebay This works for me now. Thank you

ppebay (approved):
Tested and works for me

PR #427, opened 2023-08-30
#426: Number of input files detected incorrectly

### LINKED ISSUE #426: Number of input files detected incorrectly
When I try to load balance a dataset with 40 ranks (with files named `data.$i.json` where `i` runs from `0` through `39` with no additional zero padding), it only outputs 20 json files. The logger output shows that only the first 20 input files were loaded.

When I try to load balance a dataset with 128 ranks, it only loads the first 64 files and then outputs 64 files.

With an older version of LBAF that accepts `n_ranks` in the yaml file, all files are correctly loaded and output, so the naming convention should not be an issue.

These datasets used 2 ranks per compute node. I do not know if LBAF could be reading the json header and using the number of compute nodes used instead of the number of files present.

I am not able to reproduce this incorrect behavior on the `user-defined-memory-toy-problem`.

tlamonthezie:
Fixes #426 

This PR includes:
- Fix `n_ranks `value detection by using only from the data file names (no more use `num_nodes` from data files as `n_ranks`)
- Add new optional `expected_ranks` configuration key (int value) to `from_data` configuration node. If it is set and `n_ranks` is different from `expected_ranks` a warning message is displayed to the user.

### DIFF
--- a/src/lbaf/Applications/LBAF_app.py
+++ b/src/lbaf/Applications/LBAF_app.py
@@ -95,6 +95,7 @@ class InternalParameters:
                 self.phase_ids = list(range(range_list[0], range_list[1] + 1))
             else:
                 self.phase_ids = from_data.get("phase_ids")
+            self.expected_ranks = from_data.get("expected_ranks")
 
         # Parse sampling parameters if present
         from_samplers = config.get("from_samplers")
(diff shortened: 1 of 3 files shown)


nlslatt:
@PPP This one works for me when applied on top of the install fix.

ppebay (approved):
Works also for me

PR #433, opened 2023-09-08
#428: allow swaps of empty clusters from target

### LINKED ISSUE #428: Memory-aware LB needs to allow swaps of whole clusters with empty clusters
There may be ranks that have no clusters at all or far fewer clusters than that rank has the memory to accommodate. When attempting to swap, it should be allowed to swap a whole cluster from an overloaded rank with nothing from the underloaded rank (i.e., move a whole cluster from an overloaded to an underloaded rank even if there is nothing there to swap with).

ppebay:
Fixes #428

### DIFF
--- a/src/lbaf/Execution/lbsClusteringTransferStrategy.py
+++ b/src/lbaf/Execution/lbsClusteringTransferStrategy.py
@@ -100,22 +100,28 @@ class ClusteringTransferStrategy(TransferStrategyBase):
             # Identify and perform beneficial cluster swaps
             n_swaps = 0
             for o_src in clusters_src.values():
+                # Compute source cluster size
+                sz_src = sum([o.get_load() for o in o_src])
+
+                # Iterate over potential targets to try to swap clusters
                 swapped_cluster = False
                 for r_try in targets:
-                    # Iterate over target clusters including empty set
+                    # Iterate over target clusters
                     for o_try in self.__cluster_objects(r_try).values():
                         # Decide whether swap is beneficial
-                        if self._criterion.compute(r_src, o_src, r_try, o_try) > 0.0:
+                        sz_try = sum([o.get_load() for o in o_try])
+                        c_try = self._criterion.compute(r_src, o_src, r_try, o_try)
+                        if c_try > 0.0 and c_try > 0.05 * sz_src:
                             # Perform swap
                             self._logger.info(
-                                f"Swapping cluster of size {sum([o.get_load() for o in o_src])} with cluster of size {sum([o.get_load() for o in o_try])} on {r_try.get_id()}")
+                                f"Swapping cluster of size {sz_src} with cluster of size {sz_try} on {r_try.get_id()}")
                             self._n_transfers += phase.transfer_objects(
                                 r_src, o_src, r_try, o_try)
                             swapped_cluster = True
                             n_swaps += 1
                             break
 
-                    # Break out from targets loop once one swap was performed
+                    # Break out from targets loop as soon as cluster was swapped
                     if swapped_cluster:
                         break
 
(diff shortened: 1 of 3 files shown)


ppebay:
**Key findings:**
* the initial impression that the load-balancing under-performance of the clustering-based strategy was caused by the inability to swap with empty clusters was correct;
* what was incorrect however was the first assumption that it was caused by the non-existence of a notional empty cluster on the source rank; this is because the logic being to try to offload work from overloaded to underloaded ranks, there is no point in trying to swap and empty cluster on the source in exchange for something on the trial destination; instead the symmetric exchange should be tried.

As a result, the results of the call to `__cluster_objects()` on the destination rank must always be “enriched” with the notional empty cluster, denoted as: `None: []`

ppebay:
Example of 16 iterations of the `too-complex` case, in **non-deterministic** mode:

out_00
out_16

**N.B.:** load imbalance continues to decrease and is not blocked at that level, past iteration 16.

ppebay:
Verifying that has no negative impact on other, previously well resolved, cases:
output_file_00
output_file_08

ppebay:
@nlslatt should I nonetheless merge, while you are further investigating?

nlslatt:
> @nlslatt should I nonetheless merge, while you are further investigating?

No, see Slack.

ppebay:
Results with new execution logic (now only one target allowed to be picked but with an arbitrary numbers of swaps):

Screen Shot 2023-09-09 at 4 34 19 PM

ppebay:
Further improved strategy, both for execution speed and load imbalance:
out_08

ppebay:
Results also improved for the `challenging_toy_hundreds_tasks` case (also in non-deterministic mode):

output_file_00
output_file_08

src/lbaf/Execution/lbsClusteringTransferStrategy.py
```diff
@@ -99,23 +99,33 @@ def execute(self, known_peers, phase: Phase, ave_load: float):
...
-                            break
-
-                    # Break out from targets loop once one swap was performed
-                    if swapped_cluster:
-                        break
+                        c_try = self._criterion.compute(r_src, o_src, r_try, o_try)
+                        if c_try > 0.0:
+                            # Compute source cluster size only when necessary
+                            sz_src = sum([o.get_load() for o in o_src])
+                            if  c_try > 0.05 * sz_src:
```
lifflander, line 117: Should this 5% be made into a tunable parameter?

ppebay: Yes. I am creating a follow-on issue for this.

ppebay: @lifflander @cwschilly #434

nlslatt:
@PPP This is working better for me as well

ppebay:
Thanks @nlslatt @lifflander , can we merge this PR?

src/lbaf/Execution/lbsClusteringTransferStrategy.py
```diff
@@ -99,23 +99,33 @@ def execute(self, known_peers, phase: Phase, ave_load: float):
...
+                                self._logger.info(
+                                    f"Swapping cluster {k_src} of size {sz_src} with cluster {k_try} on {r_try.get_id()}")
+                                self._n_transfers += phase.transfer_objects(
+                                    r_src, o_src, r_try, o_try)
+                                n_swaps += 1
+                                break
+
+                # In non-deterministic mode break out from targets as soon as swaps occurred
+                if n_swaps and not self._deterministic_transfer:
+                    break
```
nlslatt, line 128: Can you please explain the purpose of the two different `break` statements in this outer for loop?

nlslatt (commented):
There is also a commit message that is missing the issue number.

src/lbaf/Execution/lbsClusteringTransferStrategy.py
ppebay: Thanks for the good catch @nlslatt , lines 126--128 are indeed a leftover debug code that I have now deleted. Thanks!

nlslatt (approved):
Looks good

PR #436, opened 2023-09-14
#434: Make cluster swap relative threshold a configurable parameter

### LINKED ISSUE #434: Make cluster swap relative threshold a configurable parameter
As suggested by @lifflander in PR #433 , the 5% value in line 117 0f [Execution/lbsClusteringTransferStrategy.py](https://github.com/DARMA-tasking/LB-analysis-framework/pull/433/files/41e8215c721da7c1c729dca393b8affe55f75675#diff-52bdf682017c84ecfdd2f4caa52bf087545beaee905a7211a4ec57b9731a934c) should be made an optional (with default value of 0.05) user-defined parameter

A suggested name is `cluster_swap_rtol`. 

Please make sure that CI is updated accordingly and that this new configuration parameter is tested for invalid types.

cwschilly:
Fixes #434

### DIFF
--- a/src/lbaf/Execution/lbsClusteringTransferStrategy.py
+++ b/src/lbaf/Execution/lbsClusteringTransferStrategy.py
@@ -12,7 +12,7 @@ from ..Model.lbsPhase import Phase
 class ClusteringTransferStrategy(TransferStrategyBase):
     """A concrete class for the clustering-based transfer strategy."""
 
-    def __init__(self, criterion, parameters: dict, lgr: Logger, cluster_swap_rtol: float):
+    def __init__(self, criterion, parameters: dict, lgr: Logger):
         """Class constructor.
 
         :param criterion: a CriterionBase instance.
@@ -23,7 +23,7 @@ class ClusteringTransferStrategy(TransferStrategyBase):
         super(ClusteringTransferStrategy, self).__init__(criterion, parameters, lgr)
 
         # Initialize cluster swap relative threshold
-        self._cluster_swap_rtol = cluster_swap_rtol
+        self._cluster_swap_rtol = parameters.get("cluster_swap_rtol",0.05)
 
     def __cluster_objects(self, rank):
         """Cluster migratiable objects by shared block ID when available."""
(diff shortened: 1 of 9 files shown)


src/lbaf/Execution/lbsClusteringTransferStrategy.py
```diff
@@ -40,8 +43,9 @@ def __cluster_objects(self, rank):
             k: clusters[k]
             for k in random.sample(clusters.keys(), len(clusters))}
 
-    def __find_suitable_subclusters(self, clusters, rank_load, r_tol=0.05):
```
cwschilly, line 43: Will it be a problem that I removed this parameter? From what I could tell, it is never explicitly set (ie only the default value is used). I could of course revert back to how it was, or even set it to the new `cluster_swap_rtol`

src/lbaf/Execution/lbsTransferStrategyBase.py
```diff
@@ -147,7 +148,10 @@ def factory(
         try:
             # Instantiate and return object
             strategy = locals()[strategy_name + "TransferStrategy"]
-            return strategy(criterion, parameters, logger)
+            if strategy_name == "Clustering":
```
cwschilly, line 151: Is there a more elegant solution than this `if` block?

src/lbaf/Execution/lbsClusteringTransferStrategy.py
ppebay: `r_tol` is the value that should be passed by the configuration parameter. Whether you want to pass it, once retrieved, to this function as is currently done; or rather store it as member variable of the class, is an implementation detail I will let you decide based on what you think is most convenient.

src/lbaf/Execution/lbsTransferStrategyBase.py
ppebay: In the `develop` branch I still seem to have the following line 150:
```
 return strategy(criterion, parameters, logger)
```
at commit `0cd1502c45e683eda9963896251c7d9c87d330ac`

src/lbaf/Execution/lbsClusteringTransferStrategy.py
cwschilly: Resolved

ppebay (commented):
Is this ready for final review (it's still in draft mode)

cwschilly:
@ppebay This PR is now ready for review

ppebay (approved):
Additions to CI look good, thanks @cwschilly

PR #438, opened 2023-09-19
#437: clarify phase stepper error report

### LINKED ISSUE #437: Clarify phasestepper error report
When using the phase stepper algorithm on the `challenging_toy_fewer_tasks` problem, and specifying multiple phases in the configuration file (0, 1, 2,...), instead of reporting that there is only one phase in this problem, the following error arises:

Screenshot 2023-09-19 at 13 58 32

ppebay:
Fixes #437 

**N.B.:** the error was beyond the `PhaseStepper` only; it was due to an incorrect variable been reported when an error condition was met in the vt Reader.


pierrepebay (approved):
Looks good to me; testing it with incorrect configurations does give me more understandable error messaeges.

PR #452, opened 2023-10-04
#421: make lbsVTDataWriter at parity with the lbsVTDataReader for user-defined data

### LINKED ISSUE #421: Make lbsVTDataWriter at parity with the lbsVTDataReader for user-defined data
We want the JSON data writer to write out exactly the same data that is input by the reader.

cwschilly:
Fixes #421

### DIFF
--- a/config/user-defined-memory-toy-problem.yaml
+++ b/config/user-defined-memory-toy-problem.yaml
@@ -15,19 +15,23 @@ work_model:
     upper_bounds:
       max_memory_usage: 8000000000.0
 
+# # Specify algorithm
+# algorithm:
+#   name: InformAndTransfer
+#   phase_id: 0
+#   parameters:
+#     n_iterations: 4
+#     n_rounds: 2
+#     fanout: 2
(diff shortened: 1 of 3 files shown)


cwschilly:
Here is the current comparison between the reader (input JSON) and writer (output JSON)
image

cwschilly:
In the case of the memory-toy-problem, the index is `[rank_index, decomp_index_on_rank, task_index_on_decomp]`. I don't actually see the index being read in by the VTDataReader.

`rank_index` is output with `"home": rank_id`, and `decomp_index_on_rank` is used to calculate `shared id`,[^1] but the VTDataReader doesn't actually read it in. Similarly, I can't find any use of `task_index_on_decomp`.

It seems therefore that we may not have to write out `index`, since it isn't read in. Is this accurate @ppebay ?

[^1]:  `shared_id = decomp_index_on_rank * num_ranks + rank_index`

lifflander (approved):

ppebay (approved):

PR #451, opened 2023-09-30
#464: Improve efficiency of cluster swap stage of clustering transfer strategy

ppebay:
Resolves #464  

Related to #446

PR #460, opened 2023-10-10
#441: Run new performance profiler to see if there are places we can improve it

### LINKED ISSUE #441: Run new performance profiler to see if there are places we can improve it
https://github.com/plasma-umass/scalene

cwschilly:
Fixes #441

### DIFF
--- a/src/lbaf/Applications/LBAF_app.py
+++ b/src/lbaf/Applications/LBAF_app.py
@@ -462,10 +462,6 @@ class LBAFApplication:
                     self.__logger, phase_id, reader=reader)
                 phase.populate_from_log(phase_id)
                 phases[phase_id] = phase
-                phase = Phase(
-                    self.__logger, phase_id, reader=reader)
-                phase.populate_from_log(phase_id)
-                phases[phase_id] = phase
         else:
             n_ranks = self.__parameters.n_ranks
             phase_id = 0
@@ -523,7 +519,7 @@ class LBAFApplication:
                     self.__logger.warning(
                         "No rebalancing took place for offline load-balancing")
                 else:
-                    # Determine if a phase with same index was present
+                    # Determine if a phase with the same index was present
                     if _existing_phase := phases.get(p_id := rebalanced_phase.get_id()):
                         # Apply object timings to rebalanced phase
                         self.__logger.info(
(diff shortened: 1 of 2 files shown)


cwschilly:
While some of the "optimizations" seem trivial, others are good catches. For instance, the profiler caught this duplicated code in the `run()` function of `LBAF_app.py`:

```python
for phase_id in self.__parameters.phase_ids:
    # Create a phase and populate it
    phase = Phase(
        self.__logger, phase_id, reader=reader)
    phase.populate_from_log(phase_id)
    phases[phase_id] = phase
    phase = Phase(
        self.__logger, phase_id, reader=reader)
    phase.populate_from_log(phase_id)
    phases[phase_id] = phase
```
---
I timed the execution of the run function through multiple iterations of adding these optimizations. Averaging the times from three runs on `challenging-toy-hundreds-tasks.yaml`, here are the results:

| Change | Average `run()` time (s) |
| ----------- | ----------- |
| Original state      | 66.327 |
| Change `print_statistics` call   | 52.208 |
| Change `arrangement` sorting | 59.030 |
| Remove duplicated phase block | 38.445 |
| Rework `qoi_request` list | 45.639 |

---

Clearly some changes were more impactful than others. The timings also varied a lot, so it's not worth going by these alone. Let me know @ppebay @lifflander if you want to keep any of these changes. I'll update this table if I make any more revisions.

src/lbaf/Applications/LBAF_app.py
```diff
@@ -462,10 +462,6 @@ def run(self):
                     self.__logger, phase_id, reader=reader)
                 phase.populate_from_log(phase_id)
                 phases[phase_id] = phase
-                phase = Phase(
```
ppebay, line 465: That duplicate block is also deleted in PR #451 -- this may cause a conflict or not, we shall see

src/lbaf/Execution/lbsInformAndTransferAlgorithm.py
ppebay, line 1: line 73, `r` should be replace with `rank`

ppebay (changes requested):
Some good findings and fixes!

A couple of minor details to be changes, and we should be good to go

src/lbaf/Execution/lbsInformAndTransferAlgorithm.py
cwschilly: Resolved, thanks

cwschilly:
Thanks @ppebay , I made the requested change.

ppebay (approved):
Good findings

Issue #422, opened 2023-08-02
Stream JSON output to vt-tv


lifflander:
Explore options to stream the JSON from the lbsVTDataWriter to vt-tv.

PR #463, opened 2023-10-12
#422: Stream data to vt-tv for rendering

pierrepebay:
Fixes: #422


ppebay (approved):
Good improvements thanks

PR #470, opened 2023-10-17
#469: Improve and enrich visualization parameters

### LINKED ISSUE #469: Improve and enrich visualization parameters
The `LBAF_Viz` key should become `visualization`.

Also two new mandatory parameters should belong to the `visualization` block:
* `output_visualization_dir`
* `output_visualization_file_stem`

All existing exemplars in the repository should be made to incorporate the above, with the same values as the general output variables already present in those YAML files.

A script should be created to facilitate the transition to this new format.

cwschilly:
Fixes #469


cwschilly:
I created a script in `config/update-config-script.py`. It works by either passing a directory (`-d`) or individual config file (`-f`) on the command line:

```sh
python config/update-config-script.py -d <directory-of-config-files> -o <output-dir-for-new-config-files>
```
or
```sh
python config/update-config-script.py -f <path-to-config-file> -o <output-dir-for-new-file>
```

You can use either `-d` or `-f`, but the `-o` (output) argument is required.

If you want to overwrite existing configuration files, give their current directory as the `-o` argument.

Since I didn't use ordered dicts, the new configuration files do not have the same formatting or order of parameters as before. If this is an issue, I can fix it.

The updated config files are passed through the configuration validator.

cwschilly:
Updating the config files in `tests/unit/config` will be more complicated since some are designed to fail the validator.

nlslatt:
@cwschilly It would be nice to maintain the current order of items in the yaml files to the extent that makes sense. Otherwise, a huge diff will be generated for them even though functionally not much changes.

cwschilly:
@nlslatt @ppebay This should be good to go now. There are minor formatting changes (`True` to `true`, indentation, etc.), but using `ruamel.yaml` preserved ordering, comments, and spacing of the original files.

Is the intention for the updater script to be included within the repo? If so, I've put it in the `config` directory, but users will need to install `ruamel.yaml` before running it.

```sh
pip install ruamel.yaml
```
Alternatively, I could add it to `requirements.txt`.

ppebay (approved):
Excellent that you also used this PR to clean up the configuration files @cwschilly

ppebay:
@cwschilly you should be able to merge this one

nlslatt:
@cwschilly Looks like the script to update was moved to `scripts/update-viz-config.py` after you posted your instructions. I tried running that and had to install `ruamel.yaml`. Now I get an error because lbaf is not installed on my system. Do I really have to install it or is there another way?

cwschilly:
@nlslatt It should work now without having to install the `lbaf` package. Let me know if there are any other problems

nlslatt:
@cwschilly Thanks, that works for me now (although is surprisingly slow, more than 4 seconds for the 49-line yaml file I tested it on).

nlslatt (approved):

cwschilly:
@nlslatt I ran it with the scalene profiler and it seems like most time is spent running the ConfigurationValidator. This is just an added fail-safe, so if you need to speed things up, commenting out line 146 (`self.__validate_configuration(new_config)` may help.

PR #476, opened 2023-11-08
#475: Investigate failing PyPi package (v1.0.1)

### LINKED ISSUE #475: Investigate failing PyPi package (v1.0.1)
It seems the lbaf PyPi package is still failing, with the following ModuleNotFound error:

image

cwschilly:
Fixes #475

The issue was the use of absolute file paths; since the package directory structure is slightly different from the source code (no `src` dir), we should use relative paths whenever possible.

I also changed `overwrite` to False when lbaf is installed as a package, as discussed in the meeting.


cwschilly:
The CI failure is unrelated to this PR; it is fixed by PR #478

ppebay (approved):

PR #485, opened 2023-12-14
#484: Keep tempered algorithm in sync with implementation in vt

### LINKED ISSUE #484: Keep tempered algorithm in sync with implementation in vt
This is because [vt#2201](https://github.com/DARMA-tasking/vt/issues/2201) will result in a version of the tempered algorithm (and transfer strategies) not necessarily synchronized -- or with some improvements.

ppebay:
resolves #484

### DIFF
--- a/src/lbaf/Execution/lbsTransferStrategyBase.py
+++ b/src/lbaf/Execution/lbsTransferStrategyBase.py
@@ -65,8 +65,8 @@ class TransferStrategyBase:
         # Iterate over all provided ranks
         for r_src in ranks:
             # Ignore ranks without migratable objects
-            if not r_src.get_migratable_objects():
-                continue
+            #if not r_src.get_migratable_objects():
+            #    continue
 
             # Retrieve potential targets
             targets = known_peers.get(r_src, set()).difference({r_src})
--- a/src/lbaf/Execution/lbsTransferStrategyBase.py
+++ b/src/lbaf/Execution/lbsTransferStrategyBase.py
@@ -65,8 +65,8 @@ class TransferStrategyBase:
         # Iterate over all provided ranks
         for r_src in ranks:
             # Ignore ranks without migratable objects
-            #if not r_src.get_migratable_objects():
-            #    continue
+            if not r_src.get_migratable_objects():
+               continue
 
             # Retrieve potential targets
             targets = known_peers.get(r_src, set()).difference({r_src})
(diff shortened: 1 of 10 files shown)


cwschilly:
@ppebay This PR now passes all tests

src/lbaf/IO/lbsConfigurationValidator.py
```diff
@@ -159,7 +159,11 @@ def __init__(self, config_to_validate: dict, logger: Logger):
                          Optional("cluster_swap_rtol"): And(
                             float,
                             lambda x: x > 0.0,
-                            error="Should be of type 'float' and magnitude > 0.0"),
+                            error="Should be of type 'float' and > 0.0"),
+                         Optional("max_subclusters"): And(
+                            int,
+                            lambda x: x > 0.0,
+                            error="Should be of type 'int' and > 0"),
```
cwschilly, line 166: Should we add a unit test for this new parameter?

src/lbaf/Execution/lbsTransferStrategyBase.py
```diff
@@ -65,8 +65,8 @@ def _get_ranks_to_traverse(self, ranks: list, known_peers: list):
         # Iterate over all provided ranks
         for r_src in ranks:
             # Ignore ranks without migratable objects
-            if not r_src.get_migratable_objects():
-                continue
+            #if not r_src.get_migratable_objects():
+            #    continue
```
cwschilly, line 69: Can these lines be deleted?

cwschilly (commented):
I can write a unit test for the `max_clusters` parameter if it would be helpful--just let me know

src/lbaf/IO/lbsConfigurationValidator.py
ppebay: Yes please, that would be good

src/lbaf/Execution/lbsTransferStrategyBase.py
ppebay: Oh no -- please remove the comment `#`s
This is also a leftover debugging hack. But indeed if no objects are migratable the rank should be skipped

src/lbaf/IO/lbsConfigurationValidator.py
cwschilly: Ok, I added the tests

cwschilly (approved):
LGTM

ppebay:
@cwschilly  When undefined, the new parameter is consider to be infinite (before that addition, there was a hard-coded value of 10 in the code). When the value is not provided, it's going to be considered infinite -- which in turn might cause the code to go into a computationally intractable cluster swap exploration. 

As a result other CI tests might have come unfeasible unless the parameter is explicitly added to their configuration file.

cwschilly (approved):
The CI is hanging occasionally, but I have not been able to find anything in this PR that causes it. Any hung jobs will pass on a rerun.

PR #488, opened 2023-12-22
#486: Implement a way to optionally start subclustering after all ranks have stopped swapping

### LINKED ISSUE #486: Implement a way to optionally start subclustering after all ranks have stopped swapping
Currently the cluster swapping strategy (in non-deterministic mode) jumps into subclustering as soon as no full cluster swaps are possible on a given rank, irrespective of swapping/non-swapping on other ranks.

This has the effect of improving load balance locally but at the expense of earlier memory locality disruption than as compared to when subclustering is delayed until all ranks have exhausted full cluster swapping possibilities.

This is in contrast to the implementation of `temperedlb` in `vt`, which has some global fence between cluster swaps and the subclustering stage.

The goal of this issue is to allow for the latter (`vt`-like) option in LBAF as well.

@lifflander

ppebay:
Resolves #486

### DIFF
--- a/src/lbaf/Execution/lbsAlgorithmBase.py
+++ b/src/lbaf/Execution/lbsAlgorithmBase.py
@@ -94,8 +94,6 @@ class AlgorithmBase:
         # pylint:enable=W0641:possibly-unused-variable,C0415:import-outside-toplevel
 
         # Ensure that algorithm name is valid
-        algorithm = locals()[algorithm_name + "Algorithm"]
-        return algorithm(work_model, parameters, logger, rank_qoi, object_qoi)
         try:
             # Instantiate and return object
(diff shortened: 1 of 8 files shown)


lifflander (approved):
Looks great to me. It's nice to have it refactored out of a big function anyway.

ppebay:
Thank @lifflander I made further changes in order to be able to run the subclustering entirely **after** the swapping, governed by a new (optional) configuration parameter called `separate_subclustering`.

ppebay:
@cwschilly can you please add a new CI test for the new parameter?

ppebay:
Re-requesting review from @lifflander

ppebay:
Hello @cwschilly what's the status of this PR?

PR #499, opened 2024-01-23
#497: Add unit test to detect reader failure when communicator is absent 

### LINKED ISSUE #497: Add unit test to detect reader failure when communicator is absent
This is a follow-on issue of #493

cwschilly:
Fixes #497


ppebay (changes requested):
The `_README` and `_optimal` datasets are not used by the configuration file and can be removed from the unit test.

[**N.B.:** These are stored with the toy memory problem in the top-level `data` directory just for comparison purposes]

ppebay (approved):

PR #500, opened 2024-02-09
#498: Relax numeric conversion to float when reading JSON

### LINKED ISSUE #498: Relax numeric conversion to float when reading JSON
This should apply to at least all memory and load fields

cwschilly:
Fixes #498


lifflander (approved):
Looks good to me!

PR #508, opened 2024-05-16
#507: Deprecate LBAF-Viz and replace with calls to vt-tv 

### LINKED ISSUE #507: Deprecate LBAF-Viz and replace with calls to vt-tv
LBAF-Viz code should not be deleted yet, but should return a deprecation error.

A subsequent release will remove the deprecated code.

This is related to #410

cwschilly:
Fixes #507


ppebay (changes requested):
When using some provided configuration files (e.g. `conf.yaml`), the following error occurs as expected, but it would be better to clean up all configuration decks in the `conf` directory, to ensure that they all indeed can be used as working examples

```
  File "/Users/pppebay/Documents/Git/LB-analysis-framework/src/lbaf/Applications/LBAF_app.py", line 605, in <module>
    LBAFApplication().run()
  File "/Users/pppebay/Documents/Git/LB-analysis-framework/src/lbaf/Applications/LBAF_app.py", line 574, in run
    visualizer = Visualizer(
  File "/Users/pppebay/Documents/Git/LB-analysis-framework/src/lbaf/IO/lbsVisualizer.py", line 47, in __init__
    raise DeprecationWarning("LBAF's Visualizer has been deprecated and will be removed in a future release. Visualizations should be generated with DARMA/vt-tv.`

Of course incorrect inputs can and should remain in the CI in order to test the above failure mode.

File: config/challenging-toy-fewer-tasks.yaml
Comment (ppebay) [inline config/challenging-toy-fewer-tasks.yaml:1]: Noted removal of viz section, thanks

File: tests/unit/IO/test_vt_writer.py [hunk 1]
```diff
@@ -140,7 +140,7 @@ def test_vt_writer_required_fields_output(self):
             output_file_name = f"{output_file_prefix}.{i}.json"
             output_file = os.path.join(output_dir, output_file_name)
 
-            print(f"[{__loader__.name}] Compare input file ({input_file_name}) and output file ({output_file_name})...")
+            # print(f"[{__loader__.name}] Compare input file ({input_file_name}) and output file ({output_file_name})...")
```
ppebay, line 143: Why comment this print statement, rather than eliminate it if no longer needed?

ppebay (changes requested):
I suggest also adding an extra test in `tests/unit/config` with a configuration deck that contains, e.g., the following block:
```
visualization:
  x_ranks: 2
  y_ranks: 2
  z_ranks: 1
  object_jitter: 0.5
  rank_qoi: None
  object_qoi: None
  force_continuous_object_qoi: true
  output_visualization_dir: ../../output
  output_visualization_file_stem: output_file
```
in order to exercise the deprecation stop.

**N.B.:** this is not to replace the `test_lbs_visualizer_deprecation.py` in-code deprecation check, but **in addition** to it.

tests/unit/IO/test_vt_writer.py
cwschilly: Fixed--I removed the statement

ppebay (approved):
Additional test  is good

PR #516, opened 2024-05-24
#509:  Investigate compatibility with more recent versions of Python than 3.9

### LINKED ISSUE #509: Investigate compatibility with more recent versions of Python than 3.9 after VTK dependency removal
This issue is motivated by the following r/t error with 3.12:
```
[lbsGridStreamer] Streaming 9 load-balancing steps
zsh: segmentation fault  python LBAF_app.py -c ../../../config/conf.yaml
(base) $ python LBAF_app.py -c ../../../config/conf.yaml
```

**This is to be done only after #507 is completed**

cwschilly:
Fixes #509

This PR expands support for Python 3.10 and 3.11.

It also updates all package dependencies to the most recent supported versions. 

Related to [PR 72](https://github.com/DARMA-tasking/vt-tv/pull/72) in vt-tv

### DIFF
--- a/src/lbaf/Execution/lbsClusteringTransferStrategy.py
+++ b/src/lbaf/Execution/lbsClusteringTransferStrategy.py
@@ -61,7 +61,7 @@ class ClusteringTransferStrategy(TransferStrategyBase):
         # Return dict of computed object clusters possibly randomized
         return clusters if self._deterministic_transfer else {
             k: clusters[k]
-            for k in random.sample(clusters.keys(), len(clusters))}
+            for k in random.sample(list(clusters.keys()), len(clusters))}
 
     def __build_rank_subclusters(self, r_src: Rank) -> set:
         """Build subclusters to bring rank closest and above average load."""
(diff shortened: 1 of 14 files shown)


src/lbaf/Execution/lbsClusteringTransferStrategy.py
```diff
@@ -108,13 +108,13 @@ def __build_rank_subclusters(self, r_src: Rank) -> set:
             f"Built {len(subclusters)} subclusters from {len(clusters)} clusters in {time.time() - start_time:.3f} seconds")
         return sorted(subclusters.keys(), key=subclusters.get)
 
-    def __swap_clusters(self, phase: Phase, r_src: Rank, clusters_src:dict, targets: dict) -> int:
+    def __swap_clusters(self, phase: Phase, r_src: Rank, clusters_src:dict, targets: set) -> int:
```
ppebay, line 111: Good catch!

tox.ini
```diff
@@ -3,14 +3,16 @@
...
+    tox>=4.6.0
 skipsdist = true
-envlist = py{38,39}, pylint, coverage
+envlist = py{38,39,310,311}, pylint, coverage
 
 [gh-actions]
 python =
     3.8: py38, pylint, coverage
     3.9: py39
+    3.10: py310
```
ppebay, line 14: I did not verify it on my machine (where I have only up to 3.9.19 installed), so I am taking this as meaning that both 3.10 and 3.11 are tested in CI, correct?

I can validate that all changes made above are still working with 3.9.19.

ppebay (approved):
I confirm that the changes are still compatible with 3.9.19

I however noticed a larger number of GitHub Actions notices which, albeit not errors, should probably be addressed at some point (i.e. `Missing module docstring`).

tox.ini
cwschilly: Yes, 3.10 and 3.11 are tested in CI now as well

PR #528, opened 2024-09-11
#527: Update JSON generator to work with vt's LBDataHolder

### LINKED ISSUE #527: Update JSON generator to work with vt's `LBDataHolder`
In order for vt to successfully encode information in an object's ID, we need to:

- [ ] Add `collection_id` to each `entity` field
- [ ] Output the entire `entity` in the `communications` to/from fields (including `home`, and `migratable`)
- [ ] Use `seq_id` instead of `id` for objects/tasks (since `id` now means that the value is encoded)
- [ ] #531 

---

As a model, vt's `LBDataHolder` has this function:

```cpp
void LBDataHolder::outputEntity(nlohmann::json& j, ElementIDStruct const& id) const {
  j["type"] = "object";
  j["id"] = id.id;
  j["home"] = id.getHomeNode();
  j["migratable"] = id.isMigratable();
  if (node_idx_.find(id) != node_idx_.end()) {
    auto const& proxy_id = std::get<0>(node_idx_.find(id)->second);
    auto const& idx_vec = std::get<1>(node_idx_.find(id)->second);
    j["collection_id"] = proxy_id;
    for (std::size_t x = 0; x < idx_vec.size(); x++) {
      j["index"][x] = idx_vec[x];
    }
  } else if (node_objgroup_.find(id) != node_objgroup_.end()) {
    auto const& proxy_id = node_objgroup_.find(id)->second;
    j["objgroup_id"] = proxy_id;
  } else {
    // bare handler
  }
}
```

which is called any time an object is being written, whether in the communications field or tasks field. For example:

```cpp
      // Adding entity to tasks
      outputEntity(j["tasks"][i]["entity"], id);
      
      // Adding entity to communications["to"] or communications["from"]
      outputEntity(j["communications"][i]["from"], key.fromObj());
      outputEntity(j["communications"][i]["to"], key.toObj());
```

### LINKED ISSUE #531: Use `shared_id` instead of `shared_block_id` everywhere
In order to maintain compatibility with existing vt-generated application datasets, we should make sure that LBAF supports the `shared_id` QOI instead of `shared_block_id`.

This work will revert #525. The change should be reflected in:
- all existing JSON files
- LBAF source (e.g. `lbsObject` and `lbsRank`)
- all tests
- the JSON data files generator

I think a simple find/replace call should suffice.

This is related to #527 and should be included in PR #528

tlamonthezie:
Fixes #527
Fixes #531 

This PR provides the following changes;
1. Add packed `id` support in entities
1.2 The `Object` class -> __index renamed as `seq_id`, + addition of a `packed_id` field
2. Add `collection_id` field
2.1 Object class: Add the `collection_id` field to the attributes of the `Object` class, add `objgroup_id` field to unused params.
2.2 JSON Generator: Enable the `collection_id` field to be set by the JSON generator (JSONDataFilesMaker) (with interactive mode support)
2.3 set collection_id automatically with some default value of 7 if object is added as migratable to a rank (collection_id required by vt for migratable objects so required for the LBAF Json writer).
   _with the help of the new `TaskSpecification `typed dict defining both `time `(required) and `collection_id` (not required)_
4. VTDataWriter: copy object fields to the `entity` node inside the communication `from` or `to` nodes
5. Adapt unit tests (code, config) to work with these source updates
6. Update data sets with small entity ids by replacing `id` by `seq_id` (considering small ids sequences are in fact `seg_id`'s) regenerate to get data in communiaction endpoints nodes and add some generator config file
7. Replace `shared_block_id` by `shared_id` everywhere

Note: this PR is also needed to adapt to the new LB Data files schema defined by [VT PR 2243](https://github.com/DARMA-tasking/vt/pull/2343)

**IMPORTANT: vt-tv must be updated as well to support `seq_id` as object identifiers.
Else we have to generate entity `id` values in LBAF if  missing in input data**


data/challenging_toy_fewer_tasks/toy.0.json
nlslatt, line 1: There should not be a change from `id` to `seq_id` for files like this one that were output by vt. `seq_id` should only appear in files generated by other mechanisms, and `seq_id` should be something simple like 0, 1, 2, ... If the existing `id` is large, then it very likely was generated by vt.

nlslatt: I will clarify to say that if *any* existing `id` is large, then the file was generated by vt. vt does create a small number of small ids.

src/lbaf/IO/lbsVTDataReader.py
```diff
@@ -201,21 +202,21 @@ def _populate_rank(self, phase_id: int, rank_id: int) -> tuple:
                     # Check whether both are objects
                     if c_to.get("type") == "object" and c_from.get("type") == "object":
                         # Create receiver if it does not exist
-                        receiver_obj_id = c_to.get("id")
+                        receiver_obj_id = c_to.get("seq_id")
```
cwschilly, line 205: We want to be able to read in `"id"` as well, if it's present (i.e. from a file generated by vt, with the bit-encoded ids).

src/lbaf/IO/lbsVTDataWriter.py
```diff
@@ -62,7 +64,7 @@ def __create_tasks(self, rank_id, objects, migratable):
             task_data = {
                 "entity": {
                     "home": rank_id,
-                    "id": o.get_id(),
+                    "seq_id": o.get_id(),
```
cwschilly, line 67: Similar to the reader, if LBAF received a vt `id` from the JSON, then it should write out the `id` field here as well (and not `seq_id`).

It may help to add a flag in the `Object` class to keep track of whether or not the initial identifier is an `id` or `seq_id`.

src/lbaf/Model/lbsPhase.py
```diff
@@ -494,12 +498,16 @@ def populate_from_specification(self, spec: PhaseSpecification, multiple_sharing
             "type": "SendRecv",
             "to": {
                 "type": "object",
-                "id": comm_spec["to"]
+                "seq_id": comm_spec["to"],
```
cwschilly, line 501: Same here; the identifier key (either `id` or `seq_id`) needs to be flexible and based on the input JSON

cwschilly (changes requested):
Mainly just what @nlslatt pointed out as well.

Data files generated by `lbJSONDataFilesMaker.py` should always use `seq_id` as the object identifier.

But we still need to be able to read and write JSON generated in vt, where `id` is used to indicate a bit-packed object identifier.

src/lbaf/Model/lbsPhase.py
tlamonthezie: Ok @cwschilly there was some misunderstanding about `Use seq_id instead of id for objects/tasks (since id now means that the value is encoded)` from my side. 
I understand that LBAF should be finally be able to use either `id` or `seq_id`. This will impact the whole load balancing login as in algorithms. For example this kind of logic (BruteForceAlgorithm) will have to be modified as well
```python
object_id = objects[i].get_id()
for o in r_src.get_objects():
  if o.get_id() == object_id:
```
We have now to take care about which of `id` and `seq_id` is defined.
I will work on that.

data/challenging_toy_fewer_tasks/toy.0.json
tlamonthezie: Thanks @nlslatt for this clarification. I will consider small ids as `seq_id` and big integers as `id` and I will restore the datasets that use big ids.

src/lbaf/IO/lbsVTDataWriter.py
tlamonthezie: @cwschilly I think this is not a good idea if input files contain both id and seq_id we should keep the writer able to write the 2 values so we need to keep both in memory

src/lbaf/IO/lbsVTDataReader.py
tlamonthezie: It should be ok now

data/synthetic_lb_data/data.0.json
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+{"type":"LBDatafile","phases":[{"id":0,"tasks":[{"entity":{"seq_id":0,"home":0,"type":"object","migratable":true},"node":0,"resource":"cpu","time":1.0},{"entity":{"seq_id":1,"home":0,"type":"object","migratable":true},"node":0,"resource":"cpu","time":0.5},{"entity":{"seq_id":2,"home":0,"type":"object","migratable":true},"node":0,"resource":"cpu","time":0.5},{"entity":{"seq_id":3,"home":0,"type":"object","migratable":true},"node":0,"resource":"cpu","time":0.5}],"communications":[{"type":"SendRecv","to":{"type":"object","id":5},"messages":1,"from":{"type":"object","id":0},"bytes":2.0},{"type":"SendRecv","to":{"type":"object","id":4},"messages":1,"from":{"type":"object","id":1},"bytes":1.0},{"type":"SendRecv","to":{"type":"object","id":2},"messages":1,"from":{"type":"object","id":3},"bytes":1.0},{"type":"SendRecv","to":{"type":"object","id":8},"messages":1,"from":{"type":"object","id":3},"bytes":0.5}]}]}
```
cwschilly, line 1: The tasks field is using `seq_id` but communications still has `id`. I think it should all be `seq_id`.

data/synthetic-blocks/synthetic-dataset-blocks.0.json
```diff
@@ -1 +1 @@
...
...es":[{"id":0,"tasks":[{"entity":{"home":0,"seq_id":1,"migratable":true,"type":"object"},"node":0,"resource":"cpu","time":0.5,"user_defined":{"shared_block_id":0,"shared_bytes":9.0,"home_rank":0}},{"entity":{"home":0,"seq_id":3,"migratable":true,"type":"object"},"node":0,"resource":"cpu","time":0.5,"user_defined":{"shared_block_id":1,"shared_bytes":9.0,"home_rank":0}},{"entity":{"home":0,"seq_id":2,"migratable":true,"type":"object"},"node":0,"resource":"cpu","time":0.5,"user_defined":{"shared_block_id":1,"shared_bytes":9.0,"home_rank":0}},{"entity":{"home":0,"seq_id":0,"migratable":true,"type":"object"},"node":0,"resource":"cpu","time":1.0,"user_defined":{"shared_block_id":0,"shared_bytes":9.0,"home_rank":0}}],"communications":[{"type":"SendRecv","to":{"type":"object","seq_id":5},"messages":1,"from":{"type":"object","seq_id":0},"bytes":2.0},{"type":"SendRecv","to":{"type":"object","seq_id":4},"messages":1,"from":{"type":"object","seq_id":1},"bytes":1.0},{"type":"SendRecv","to":{"type":"object","seq_id":2},"messages":1,"from":{"type":"object","seq_id":3},"bytes":1.0},{"type":"SendRecv","to":{"type":"object","seq_id":8},"messages":1,"from":{"type":"object","seq_id":3},"bytes":0.5}]}]}
```
cwschilly, line 1: I see that you updated the writer to add the entire `entity` field to the `communications`, but it doesn't look like the JSON files have been updated with this change.

This is only strictly required for datasets that we intend to also test in `vt`. For now, that's only `synthetic-blocks`, but I also think `ccm_example` might be relevant down the line.

Could you please regenerate the data files for those two problems?

tlamonthezie: you are right the writer is OK but data is not. Its because I only renamed id to seq id in the data sets without regenerating it using the generator. I will update these data.

tlamonthezie: It seems that only `synthetic-blocks` dataset contains its YAML specification.
We have to either edit manually or add a yaml file to `synthetic_lb_data` and `ccm_example`

cwschilly: I think `synthetic-blocks` is the highest priority, since that's the test case we're using across the DARMA-tasking repos.

I'm not sure how the other two datasets were generated, but since there are no immediate plans to use them in vt, it might be okay to leave them as they are.

tlamonthezie: It has been fixed for the 3 examples

data/synthetic-blocks/synthetic-dataset-blocks.0.json
```diff
@@ -1 +1 @@
...
..."migratable":true,"seq_id":1,"type":"object"},"node":0,"resource":"cpu","time":0.5,"user_defined":{"home_rank":0,"shared_block_id":0,"shared_bytes":9.0}},{"entity":{"home":0,"migratable":true,"seq_id":2,"type":"object"},"node":0,"resource":"cpu","time":0.5,"user_defined":{"home_rank":0,"shared_block_id":1,"shared_bytes":9.0}},{"entity":{"home":0,"migratable":true,"seq_id":3,"type":"object"},"node":0,"resource":"cpu","time":0.5,"user_defined":{"home_rank":0,"shared_block_id":1,"shared_bytes":9.0}}],"communications":[{"bytes":2.0,"from":{"home":0,"migratable":true,"seq_id":0,"type":"object"},"messages":1,"type":"SendRecv","to":{"home":1,"migratable":true,"seq_id":5,"type":"object"}},{"bytes":1.0,"from":{"home":0,"migratable":true,"seq_id":1,"type":"object"},"messages":1,"type":"SendRecv","to":{"home":1,"migratable":true,"seq_id":4,"type":"object"}},{"bytes":1.0,"from":{"home":0,"migratable":true,"seq_id":3,"type":"object"},"messages":1,"type":"SendRecv","to":{"home":0,"migratable":true,"seq_id":2,"type":"object"}},{"bytes":0.5,"from":{"home":0,"migratable":true,"seq_id":3,"type":"object"},"messages":1,"type":"SendRecv","to":{"home":2,"migratable":true,"seq_id":8,"type":"object"}}]}]}
```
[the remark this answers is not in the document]
tlamonthezie: Done

data/ccm_example/ccm_example.yaml
ppebay, line 1: Is this "CCM example" connected to any particular example in the CCM-MILP repo?

data/synthetic_lb_data/synthetic_lb_data.yaml
ppebay, line 1: Please add comment on top that this is meant to be the same case as CCM-MILP"s `examples/data/synthetic_blocks.py`

scripts/copy_dataset.py
```diff
@@ -0,0 +1,72 @@
...
+                        decompressed_dict = json.loads(compr_bytes.decode("utf-8"))
+                    except json.JSONDecodeError:
+                        print(f"Invalid JSON file: {i.name}. Ignored.")
+
+            if decompressed_dict is None:
+                continue
+
+            json_str = json.dumps(decompressed_dict, separators=(',', ':'))
+
+            # Issue #527: replace `id` by seq_id in tasks entities and communication nodes
```
ppebay, line 49: Why is this commented out? If not needed it should be removed.

tests/unit/IO/test_vt_writer.py
```diff
@@ -38,25 +38,35 @@ def __list_optional_keys_recursive(self, nodes: Any, dot_path: str = ''):
             return opt_nodes
 
         # Key & node
-        if isinstance(nodes, dict):
+        elif isinstance(nodes, dict):
```
ppebay, line 41: These nested conditionals are hard to follow/understand. Please comment this code to express intent

src/lbaf/Model/lbsObject.py
```diff
@@ -3,25 +3,49 @@
...
-
 class Object:
-    """A class representing an object with load and communicator"""
+    """A class representing an object with load and communicator
+
+    Constructor arguments:
+    :arg seq_id: small int identifier, commonly the task index, defaults to None
+    :arg packed_id: alt identifier (generated by vt as a bitpacked value), defaults to None
+    :arg r_id: rank ID, defaults to None
+    :arg load: the load, defaults to 0.0
```
ppebay, line 13: replace with `load: the computational time, also known as load, defaults to 0.0`

ppebay (changes requested):

data/ccm_example/ccm_example.yaml
tlamonthezie: I think that there is no connection.

data/synthetic_lb_data/synthetic_lb_data.yaml
tlamonthezie: Done

scripts/copy_dataset.py
tlamonthezie: Lines have been removed

tests/unit/IO/test_vt_writer.py
tlamonthezie: Done

src/lbaf/Model/lbsObject.py
tlamonthezie: Done

data/challenging_toy_fewer_tasks/toy.0.json
tlamonthezie: Done

src/lbaf/Model/lbsPhase.py
tlamonthezie: Done

data/synthetic_lb_data/data.0.json
tlamonthezie: Fixed

data/synthetic-blocks/synthetic-dataset-blocks.0.json
tlamonthezie: Fixed

ppebay (approved):
Thanks for changes @tlamonthezie

ppebay:
Requesting now a SNL review @nlslatt @lifflander due to connection with vt

tlamonthezie:
#531 now included in the PR

src/lbaf/Model/lbsObject.py
```diff
@@ -3,25 +3,49 @@
...
         size: float=0.0,
         comm: Optional[ObjectCommunicator]=None,
         user_defined: dict=None,
-        subphases: list=None):
+        subphases: list=None,
+        collection_id: Optional[int] = None):
+
+        # Check that id is provided as defined in LBDatafile schema
+        if packed_id is None and seq_id is None:
+            raise ValueError('Either packed_id (bit-encoded ID) or i (seq ID) must be provided.')
```
cwschilly, line 35: I think this might be a typo: `... or i (seq ID) ...`

src/lbaf/Model/lbsObject.py
```diff
@@ -83,11 +110,27 @@ def __init__(
             raise TypeError(f"subphases: {subphases} is of type {type(subphases)} but must be <class 'list'>")
 
     def __repr__(self):
-        return f"Object id: {self.__index}, load: {self.__load}"
+        return f"Object id: {self.__seq_id}, load: {self.__load}"
```
cwschilly, line 113: Could this use the `get_id()` function you wrote (that handles either type of id)?

```python
def __repr__(self):
    return f"Object ID: {get_id()}, load: {self.__load}"
```

cwschilly (commented):
The newly generated `synthetic-blocks` data works with vt :+1:

src/lbaf/Model/lbsObject.py
tlamonthezie: Fixed

src/lbaf/Model/lbsObject.py
tlamonthezie: This is done

cwschilly (approved):
LGTM (and works with vt tests)

scripts/copy_dataset.py
nlslatt, line 1: Why is there a python script to "copy" json files by reading, decompressing, converting to a dict, converting back to a string, writing, and compressing? This seems very inefficient.

tlamonthezie: Hello @nlslatt this cript was in fact very helpful to resolve this issue and might be in the future.
It enables for example to de-compress dataset, then you can update the dataset using a text editor, then call the script again to re-compress.
It was particularly helpful when renaming json keys for this issue in compressed datasets.

lifflander (approved):

PR #538, opened 2024-09-20
#537: Fix broken CI

### LINKED ISSUE #537: Fix broken CI
Looks like LBAF's CI is broken (see [this run](https://github.com/DARMA-tasking/LB-analysis-framework/actions/runs/10964079320/job/30446954826) on the latest push to `develop`).

We need to
- fix the failures (likely related to the changes in acceptable schema in vt)
- figure out why PR #528 was passing CI before it was merged

cwschilly:
Fixes #537


cwschilly:
The CI is fixed but I'm still investigating why it passed on PR #528 . I've rerun the pipelines on that branch and they're failing now.

Could it be related to this block in `src/lbaf/__init__.py`?

```python
    # If using the lbaf package, don't overwrite
    if importlib.util.find_spec('lbaf'):
        JSONDataFilesValidatorLoader().run(overwrite=False)
    else:
        JSONDataFilesValidatorLoader().run(overwrite=True)
```

tlamonthezie:
> The CI is fixed now but I'm still investigating why all tests passed on PR #528 . I've rerun the pipelines on that branch and they're failing now.
> 
> @tlamonthezie Do you have any idea what would have made the tests pass originally?

This is probably because the schema changed in vt in LBDatafile_schema.py with the recent addition of validate_ids constraint. 
In lbaf tests we tests some invalid schemas but the expected first error was no more thrown because new error was thrown first in the wrong schema datasets (because validate_ids fail and there are then 2 errors in thoses schemas). 
Any update in the vt schema in fact might make lbaf schema tests fail even if it previously succeeded.
In this particular case resolving the data files errors thrown by validate_ids should make ci passing again because the remaining error will be the expected one again.

tlamonthezie:
Hello @cwschilly it seems that the bug is in VT in the schema.

The recently added condidion `if field['migratable'] and 'seq_id' in field and 'collection_id' not in field:` is definitely the error cause.
It is not testing if the optional `migratable` key is present. So if key is not present (in a communication  `from` or `to` node) we get a failure of type `KeyError('migratable')` raised when `validate_ids` is called.
To fix it we should fix VT schema by replacing `if field['migratable']` by `if field.get('migratable') is True`

tlamonthezie (approved):
Looks good for current PR tests passing again.
But initial error from vt has also to be fixed as described in another PR comment. (`LBDatafile_schema.py ` -> `if field['migratable'] and 'seq_id' in field and 'collection_id' not in field` generates an error if the optional field `migratable `is not defined)

PR #545, opened 2024-09-24
#544: Add a new algorithm to implement the permutations prescribed by CCM-MILP

### LINKED ISSUE #544: Add a new algorithm to implement the permutations prescribed by CCM MILP
This is somewhat similar to what the permutation code does in CCM-MILP, except that what we want here is not to modify the JSON input data, but rather to modify the loaded phase based on the prescribed permutation, in particular so it may be directly fed into vt-tv and rendered.

ppebay:
Resolves #544

### DIFF
--- a/src/lbaf/Execution/lbsAlgorithmBase.py
+++ b/src/lbaf/Execution/lbsAlgorithmBase.py
@@ -103,8 +103,6 @@ class AlgorithmBase:
         # pylint:enable=W0641:possibly-unused-variable,C0415:import-outside-toplevel
 
         # Ensure that algorithm name is valid
-        algorithm = locals()[algorithm_name + "Algorithm"]
-        return algorithm(work_model, parameters, logger, rank_qoi, object_qoi)
         try:
             # Instantiate and return object
             algorithm = locals()[algorithm_name + "Algorithm"]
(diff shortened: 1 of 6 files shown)


src/lbaf/Execution/lbsAlgorithmBase.py
```diff
@@ -97,11 +97,14 @@ def factory(
...
         from .lbsInformAndTransferAlgorithm import InformAndTransferAlgorithm
         from .lbsBruteForceAlgorithm import BruteForceAlgorithm
+        from .lbsPrescribedPermutationAlgorithm import PrescribedPermutationAlgorithm
         from .lbsPhaseStepperAlgorithm import PhaseStepperAlgorithm
         from .lbsCentralizedPrefixOptimizerAlgorithm import CentralizedPrefixOptimizerAlgorithm
         # pylint:enable=W0641:possibly-unused-variable,C0415:import-outside-toplevel
 
         # Ensure that algorithm name is valid
+        algorithm = locals()[algorithm_name + "Algorithm"]
+        return algorithm(work_model, parameters, logger, rank_qoi, object_qoi)
```
cwschilly, line 107: These lines were most likely for debugging and can be removed now

tlamonthezie (approved):
Nothing special to report.
Runs successfully using the new `PrescribedPermutationAlgorithm` and with the new `permute.yaml` config

cwschilly (commented):
Looks good, but should we add a test for this algorithm?

Issue #543, opened 2024-09-24
Remove all `vtk` and `LBAF_Viz` code from LBAF


cwschilly:
Now that vt-tv has fully taken over visualization, we can safely remove the old code.

EDIT: Before removal, make sure that we can specify which phase(s) we want vt-tv to visualize. For now it's only creating the PNG for the initial phase.

PR #546, opened 2024-09-24
#542: Fix as many pylint issues as possible before release

### LINKED ISSUE #542: Fix as many pylint issues as possible before release
Also, as part of this issue, it could be useful to check the `disable` statements for PyLint that are interspersed amongst the code base, as some may have become obsolete and/or irrelevant.

cwschilly:
Fixes #542 

These are all "easy" fixes; warnings that would require more refactoring of the code (e.g. `too-many-*`) have not been resolved.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -3,7 +3,7 @@
 # A comma-separated list of package or module names from where C extensions may
 # be loaded. Extensions are loading into the active Python interpreter and may
 # run arbitrary code.
-extension-pkg-allow-list=
+extension-pkg-allow-list=vttv
 
 # A comma-separated list of package or module names from where C extensions may
 # be loaded. Extensions are loading into the active Python interpreter and may
--- a/.pylintrc
+++ b/.pylintrc
@@ -92,7 +92,8 @@ disable=raw-checker-failed,
         logging-fstring-interpolation,
         missing-module-docstring,
         missing-class-docstring,
-        protected-access
+        protected-access,
+        too-few-public-methods
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option
@@ -515,7 +516,7 @@ max-branches=12
 max-locals=15
 
 # Maximum number of parents for a class (see R0901).
-max-parents=7
+max-parents=8
 
 # Maximum number of public methods for a class (see R0904).
 max-public-methods=20
--- a/.pylintrc
+++ b/.pylintrc
@@ -216,6 +216,8 @@ good-names=i, # loop variable
            s,
            e, # exception
            Run,
+           A,
+           B,
            _
 
 # Good variable names regexes, separated by a comma. If names match any regex,
--- a/src/lbaf/Applications/LBAF_app.py
+++ b/src/lbaf/Applications/LBAF_app.py
@@ -50,9 +50,9 @@ import yaml
 
 try:
     import vttv
-    using_vttv = True
+    USING_VTTV = True
 except ModuleNotFoundError:
-    using_vttv = False
+    USING_VTTV = False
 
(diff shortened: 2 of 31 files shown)


tlamonthezie (changes requested):
Seems to be good.

However some pylint warnings could be still removed for `lbsTemperedCriterion`, `lbsPrescribedPermutationAlgorithm`, `lbsVisualizer`

Here is my output with only errors or warnings verifications:
```shell
$ pylint --rcfile=.pylintrc --disable=R,C src
************* Module lbaf.Execution.lbsTemperedCriterion
src/lbaf/Execution/lbsTemperedCriterion.py:8:0: W0223: Method 'estimate' is abstract in class 'CriterionBase' but is not overridden in child class 'TemperedCriterion' (abstract-method)
************* Module lbaf.Execution.lbsPrescribedPermutationAlgorithm
src/lbaf/Execution/lbsPrescribedPermutationAlgorithm.py:80:12: W1309: Using an f-string that does not have any interpolated variables (f-string-without-interpolation)
src/lbaf/Execution/lbsPrescribedPermutationAlgorithm.py:77:8: W0612: Unused variable 'stats' (unused-variable)
src/lbaf/Execution/lbsPrescribedPermutationAlgorithm.py:1:0: W0611: Unused import random (unused-import)
src/lbaf/Execution/lbsPrescribedPermutationAlgorithm.py:2:0: W0611: Unused import time (unused-import)
src/lbaf/Execution/lbsPrescribedPermutationAlgorithm.py:6:0: W0611: Unused Rank imported from Model.lbsRank (unused-import)
src/lbaf/Execution/lbsPrescribedPermutationAlgorithm.py:7:0: W0611: Unused Message imported from Model.lbsMessage (unused-import)
************* Module lbaf.IO.lbsVisualizer
src/lbaf/IO/lbsVisualizer.py:53:8: W0101: Unreachable code (unreachable)
```

Note: the last one will be removed by PR #543

cwschilly:
Thanks @tlamonthezie , I fixed the warnings in `lbsPrescribedPermutationAlgorithm.py`.

I left the `lbsVisualizer` as it is because (like you said) it'll be removed very soon. 

As for the `estimate` warning, it looks like `CriterionBase` defines this as a method, but it's not clear to me whether it's really used.

`TemperedLB` doesn't implement it at all (which causes the warning), while `StrictLocalizingCriterion` implements it by simply forwarding it to the compute function:

```python
def estimate(self, r_src: Rank, o_src: list, *args) -> float:
        """Estimate is compute because all information is local for this criterion."""
        return self.compute(r_src, o_src, *args)
```

It seems then that we could just remove the `estimate` function entirely. I mentioned this to @ppebay -- I think this will require further investigation outside of this PR.

ppebay:
> Thanks @tlamonthezie , I fixed the warnings in `lbsPrescribedPermutationAlgorithm.py`.
> 
> I left the `lbsVisualizer` as it is because (like you said) it'll be removed very soon.
> 
> As for the `estimate` warning, it looks like `CriterionBase` defines this as a method, but it's not clear to me whether it's really used.
> 
> * `TemperedCriterion` doesn't implement it at all (which causes the warning)
> * `StrictLocalizingCriterion` implements it by simply forwarding it to the compute function:
> 
> ```python
> def estimate(self, r_src: Rank, o_src: list, *args) -> float:
>         """Estimate is compute because all information is local for this criterion."""
>         return self.compute(r_src, o_src, *args)
> ```
> 
> It seems then that we could just remove the `estimate` function entirely. I mentioned this to @ppebay -- I think this will require further investigation outside of this PR.

Yes, please create follow-on issue @cwschilly and let's leave it in the code for now. Thanks

ppebay (approved):

tlamonthezie (approved):

PR #549, opened 2024-09-26
#543: Remove all `vtk` and `LBAF_Viz` code from LBAF

### LINKED ISSUE #543: Remove all `vtk` and `LBAF_Viz` code from LBAF

cwschilly:
Fixes #543 

Removes `lbsVisualizer`, `lbsGridStreamer`, and `MoveCountsViewer`.


src/lbaf/Applications/MoveCountsViewer.py
tlamonthezie, line 1: Just a question. Even if we remove VTK related code: is ther some scenario for which this script could still be useful ?
(if yes it might be moved to another repository if it is the case, or rewritten in C++ inside the vt-tv repository or any other option and then we might have to create a new issue)

@ppebay is that script really no more used ?

tlamonthezie (commented):
Nothin special to report except a question

src/lbaf/Applications/MoveCountsViewer.py
ppebay: I agree that moving/rewriting in C++ would be the best option (in conjunction with vt-tv)

ppebay (approved):

src/lbaf/Applications/MoveCountsViewer.py
cwschilly: Issue created in vt-tv: [vttv#124](https://github.com/DARMA-tasking/vt-tv/issues/124)

PR #559, opened 2024-10-21
558 Add optional node-level memory constraint instead of rank-level constraint to loosen

### LINKED ISSUE #558: Add node-level memory constraint

lifflander:
Fixes #558

### DIFF
--- a/src/lbaf/Execution/lbsClusteringTransferStrategy.py
+++ b/src/lbaf/Execution/lbsClusteringTransferStrategy.py
@@ -132,8 +132,8 @@ class ClusteringTransferStrategy(TransferStrategyBase):
 
         subclusters = {}
 
-        amount_over_average = src_load - ave_load;
-        amount_under_average = ave_load - try_load;
+        amount_over_average = src_load - ave_load
+        amount_under_average = ave_load - try_load
 
         # print(f"amount_over_average={amount_over_average}, amount_under_average={amount_under_average}")
 
--- a/src/lbaf/Execution/lbsClusteringTransferStrategy.py
+++ b/src/lbaf/Execution/lbsClusteringTransferStrategy.py
@@ -68,6 +68,7 @@ class ClusteringTransferStrategy(TransferStrategyBase):
 
         # Determine whether subclustering is performed immediately after swapping
         self.__separate_subclustering = parameters.get("separate_subclustering", False)
+        self.__disable_subclustering = parameters.get("subclustering_disabled", False)
         self._logger.info(
             f"Enter subclustering immediately after cluster swapping: {self.__separate_subclustering}")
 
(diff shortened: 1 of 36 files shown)


ppebay:
@lifflander we seem to have a confusion between the two types of object ids/indices now:
```
[LBAF_app] Writing single phase 0 to JSON files
multiprocessing.pool.RemoteTraceback: 
"""
  File "/Users/pppebay/miniconda3/envs/lbaf39/lib/python3.9/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/Users/pppebay/Documents/Git/LB-analysis-framework/src/lbaf/IO/lbsVTDataWriter.py", line 279, in _json_writer
    serial_json = self._json_serializer(rank_phases_double)
  File "/Users/pppebay/Documents/Git/LB-analysis-framework/src/lbaf/IO/lbsVTDataWriter.py", line 251, in _json_serializer
    self.__create_tasks(
  File "/Users/pppebay/Documents/Git/LB-analysis-framework/src/lbaf/IO/lbsVTDataWriter.py", line 126, in __create_tasks
    if o.get_index() is not None:
  File "/Users/pppebay/Documents/Git/LB-analysis-framework/src/lbaf/Model/lbsObject.py", line 184, in get_index
    return self.__index
AttributeError: 'Object' object has no attribute '_Object__index'
"""

The above exception was the direct cause of the following exception:

  File "/Users/pppebay/Documents/Git/LB-analysis-framework/src/lbaf/Applications/LBAF_app.py", line 693, in <module>
    LBAFApplication().run()
  File "/Users/pppebay/Documents/Git/LB-analysis-framework/src/lbaf/Applications/LBAF_app.py", line 623, in run
    self.__json_writer.write(
  File "/Users/pppebay/Documents/Git/LB-analysis-framework/src/lbaf/IO/lbsVTDataWriter.py", line 316, in write
    for file_name in results:
  File "/Users/pppebay/miniconda3/envs/lbaf39/lib/python3.9/multiprocessing/pool.py", line 870, in next
    raise value
AttributeError: 'Object' object has no attribute '_Object__index'
```

This is because there is still a method called `get_index()` in `lbsObject`, which is used by the data writer.

Wasn't the intent of your new `collection_id` to supersede the object `index` (rather than be an addition to it)?

For instance if I modify the current `get_index()` in the development branch as follows:
```
    def get_index(self) -> Optional[list]:
        """Return the object's index."""
        return self.get_id()
```
then the runtime error above disappears.

In summary: should `index` be entirely eliminated? The problem I am seeing with this is that the JSON files contain `index` keys for the objects.

ppebay:
Assigning @cwschilly to resolve the issues caused by the recent merges Tx

src/lbaf/Execution/lbsClusteringTransferStrategy.py
```diff
@@ -116,6 +118,104 @@ def __build_rank_clusters(self, rank: Rank, with_nullset) -> dict:
...
+                    # Additional filters prior to subclustering
+                    if c_try <= self.__subclustering_minimum_improvement * r_src.get_load() or \
+                        r_src.get_load() < self.__subclustering_threshold * max_load:
+                        continue
+
+                    l_try = abs(r_try.get_load() + objects_load - ave_load)
+                    if l_try < l_dst:
+                        c_dst, r_dst, l_dst = c_try, r_try, l_try
+                    elif l_try == l_dst and c_try > c_dst:
+                        c_dst, r_dst = c_try, r_try
```
cwschilly, line 204: How should we handle the `else` condition of this block? (There is a pylint error because `r_dst` might not be defined)

lifflander:
Supersceded by #581

PR #569, opened 2024-12-13
#567: QOIs: refactor to use a decorator

### LINKED ISSUE #567: Improve QOI detection
I am not a fan of this:

```python
        r_qoi_list = ["work"]
        with open(os.path.join(target_dir, rank_script_name), 'r', encoding="utf-8") as f:
            lines = f.readlines()
            for line in lines:
                if line[8:12] == "get_":
                    r_qoi_list.append(line[12:line.find("(")])

        # Create list of all Object QOI (Object.get_*)
        o_qoi_list = []
        with open(os.path.join(target_dir, object_script_name), 'r', encoding="utf-8") as f:
            lines = f.readlines()
            for line in lines:
                if line[8:12] == "get_":
                    o_qoi_list.append(line[12:line.find("(")])
```

Loading the text of a code file and then using a regex to search a file to match a function name is not the way to solve this.

The plan is to re-implement this cleanly with decorators and allow the user to specify exactly which functions should be QOIs.

lifflander:
Fixes #567

### DIFF
--- a/src/lbaf/Applications/LBAF_app.py
+++ b/src/lbaf/Applications/LBAF_app.py
@@ -48,6 +48,9 @@ from typing import Any, Dict, Optional, cast
 import importlib
 import yaml
 
+from ..Model.lbsRank import Rank
+from ..Model.lbsObject import Object
+
 try:
     import vttv
     USING_VTTV = True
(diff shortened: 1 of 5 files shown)


ppebay (approved):
I like your use of the `@qoi` decorator but as I had to resolve a conflict with the new develop following the merger of PR #568 can you please have a look @lifflander at the resolved merged version of pR #569 ? Tx

lifflander:
Looks good to me, I just rebased it so we don't have a merge commit in there.

PR #574, opened 2024-12-20
#571: Resolve problem of shallow copying of objects that invalidates lookup table

### LINKED ISSUE #571: Resolve problem of shallow copying of objects that invalidates lookup table
This is related to the bug uncovered as part of #566 PR #570.

The main problem is that `get_rank_id()` on an `Object` instance behaves erratically once said object is shallow copied from one `Rank` instance to another. 

This has been remedied in the VT data writer by causing the lookup table to be recomputed each time the rank owner of an object is needed. 

However the `get_rank_id()` method should either be corrected (which would imply no longer shallow copying objects) or removed entirely. This is to be discussed in particular with @lifflander

ppebay:
Fixes #571 

**N.B.:** This requires a slightly different handling of shared block deletion.

The only place where this change actually has an impact currently is in the vt data reader, specifically lines 191 and 211

### DIFF
--- a/src/lbaf/IO/lbsVTDataWriter.py
+++ b/src/lbaf/IO/lbsVTDataWriter.py
@@ -188,7 +188,8 @@ class VTDataWriter:
                 if len(sender_obj) == 1:
                     # Retrieve communications with single sender
                     sender_obj = sender_obj[0]
-                    sender_rank_id = self.__find_object_rank(phase, sender_obj).get_id()
+                    sender_rank_id = sender_obj.get_rank_id()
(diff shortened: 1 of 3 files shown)


ppebay:
@lifflander this alternative approach (with more deep copies but therefore more "iteration persistence") is in my opinion better in the context of this **analysis framework**, where performance is not the main focus (but keeping as many options open as possible, by retaining intermediate states, is). 

If you agree we can merge this alternative approach.

lifflander (approved):
Yes, I think a deep copy is fine. If we notice performance problems, we can go from there.

PR #581, opened 2024-12-30
#558: Restructured PR to add node level memory constraint post recent changes

### LINKED ISSUE #558: Add node-level memory constraint

ppebay:
Resolves #558

### DIFF
--- a/config/synthetic-blocks.yaml
+++ b/config/synthetic-blocks.yaml
@@ -1,6 +1,7 @@
 # Specify input
 from_data:
   data_stem: ../data/synthetic-blocks/synthetic-dataset-blocks
+  ranks_per_node: 2
   phase_ids:
   - 0
 check_schema: false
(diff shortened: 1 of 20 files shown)


ppebay:
@lifflander this is an interesting case (`synthetic-blocks`, load-only + memory constraint):

### Configuration 1:
* 2 ranks per node
* memory constraint of 36 (i.e. maximum of 4 blocks per **node**)
We obtain an imbalance of 0, in the following steps:

synthetic-dataset-blocks0
synthetic-dataset-blocks1
synthetic-dataset-blocks2
synthetic-dataset-blocks3

What is interesting to observe is that the final state not only each node has a maximum of 4 blocks, but also each rank has at most 2 blocks  (memory = 36.0/18.0):

Screenshot 2025-01-16 at 02 52 13

cwschilly (approved):
LGTM

lifflander (approved):
Looks good to me!

PR #592, opened 2025-02-27
#591: Revamp the JSON file makes to add functionality

### LINKED ISSUE #591: Revamp the JSON file maker to add functionality
1. Modify the `load_sample` routine to become `load_specification` and have it read a Python file spec, rather than use a hard-coded "sample" as it's doing now

cwschilly:
Fixes #591


cwschilly:
The spec file generator is implemented only for a single phase.

The following code produces the subsequent YAML spec:

```python
from lbsJSONSpecFileMaker import JSONSpecFileMaker

def main():
    user_defined_t1 = {
        "xxxx": "yyyy"
    }

    spec = JSONSpecFileMaker()

    t1 = spec.createTask(time=0.5, seq_id=10, user_defined=user_defined_t1)
    t2 = spec.createTask(time=0.7)

    s1 = spec.createSharedBlock(bytes=10000.0, tasks=[t1, t2])

    r1 = spec.createRank(id=0)
    spec.assignTask(t1, r1)
    spec.assignTask(t2, r1)

    spec.write()

main()
```

YAML spec:

```yaml
communications: {}
ranks:
  0:
    tasks:
    - 0
    - 10
shared_blocks:
  0:
    home_rank: 0
    size: 10000.0
    tasks:
    - 10
    - 0
tasks:
  0:
    collection_id: 7
    home: 0
    migratable: true
    node: 0
    seq_id: 0
    time: 0.7
  10:
    collection_id: 7
    home: 0
    migratable: true
    node: 0
    seq_id: 10
    time: 0.5
    user_defined:
      xxxx: yyyy

```

ppebay:
@cwschilly this PR is still marked as WIP (Draft), is it ready to be reviewed nonetheless?

src/lbaf/Utils/lbsJSONSpecFileMaker.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+
+    ###########################################################################
+    ## Public Functions for creating JSON fields
+
+    def createTask(self,
+            time: float,
+            seq_id: int = -1,
+            collection_id: int = 7,
```
ppebay, line 85: Why 7?

src/lbaf/Utils/lbsJSONSpecFileMaker.py
```diff
@@ -0,0 +1,215 @@
...
+        if user_defined is not None:
+            self.ranks[id]["user_defined"] = user_defined
+
+        return self.ranks[id]
+
+    def assignTask(self,
+            task: Union[int, TaskSpecification],
+            rank: Union[int, RankSpecification]
+        ) -> None:
+        task_id = task if isinstance(task, int) else task["seq_id"]
```
ppebay, line 156: A comment would be welcome here, I am not sure to understand the intent

src/lbaf/Utils/lbsJSONSpecFileMaker.py
```diff
@@ -0,0 +1,215 @@
...
+
+    ###########################################################################
+    ## Private functions for assertions
+
+    def assertIDExists_(self, test_id: int, id_type: str):
+        assert test_id in self.id_sets[id_type], \
+               f"Task {test_id} has not been created yet. Use createTask() method."
+
+    def assertAllTasksHaveBeenAssigned_(self):
+        for t in self.tasks.keys():
```
ppebay, line 72: Can't you just iterate over `self.tasks` here?

ppebay (changes requested):

src/lbaf/Utils/lbsJSONSpecFileMaker.py
cwschilly: In vt-generated JSON, `collection_id` is often set to 7, but I'm not sure why. @lifflander and @nlslatt , is this a valid default value?

lifflander: This just happens to be the first collection ID that VT assigns to the first collection it creates. If we are going to generate a collection ID with this (which is probably unnecessary) we can start at 1.

ppebay (approved):
Now it's good

PR #595, opened 2025-03-14
#594: Make alpha a parameterized constant on the rank

### LINKED ISSUE #594: Make alpha a parameterized constant on the rank
We want a different alpha per rank that ranges from 0 to infinity. This value will be read from the json metadata in the header, as the value `rank_alpha`. The load balancer should then use this to scale the compute time as it computes the work for that rank.

```json
{
  "metadata": {
    "phases": {
      "identical_to_previous": {
        "list": [],
        "range": []
      },
      "skipped": {
        "list": [],
        "range": []
      }
    },
    "rank": 0,
    "shared_node": {
      "id": 0,
      "num_nodes": 1,
      "rank": 0,
      "size": 1
    },
    "rank_alpha": 1.1,
    "type": "LBDatafile"
  }
}
```

ppebay:
Resolves #594


ppebay:
Rank-local alpha appears to be working, i.e. in the case of communications-only (all rank alphas set to null in JSON files):

Screenshot 2025-03-15 at 11 35 56

@cwschilly @lifflander CI is still failing and will continue doing so until the (imported) JSON validator in VT is updated to integrate the new `"rank_alpha"` key: [vt#2409](https://github.com/DARMA-tasking/vt/issues/2409)

config/ccm-example.yaml
```diff
@@ -9,7 +9,6 @@ check_schema: false
 work_model:
   name: AffineCombination
   parameters:
-    alpha: 1.0
```
lifflander, line 12: Should these have a new variable in the yaml, like `default_alpha` when it's not specified?

lifflander:
This is looking good to me!

config/ccm-example.yaml
ppebay: Oh this one had been forgotten! Removing it right now. Thanks.

ppebay:
Now ready for review @lifflander @cwschilly

src/lbaf/Applications/LBAF_app.py
```diff
@@ -557,11 +557,11 @@ def run(self, cfg=None, cfg_dir=None):
...
             self.__logger.info("Starting brute force optimization")
             objects = initial_phase.get_objects()
-            alpha, beta, gamma = [
+            beta, gamma = [
                 self.__parameters.work_model.get("parameters", {}).get(k)
-                for k in ("alpha", "beta", "gamma")]
+                for k in ("beta", "gamma")]
             _n_a, _w_min_max, a_min_max = lbstats.compute_min_max_arrangements_work(
-                objects, alpha, beta, gamma, n_ranks, logger=self.__logger)
+                objects, 1.0, beta, gamma, n_ranks, logger=self.__logger)
```
cwschilly, line 564: Hard-coded `1.0`--perhaps we could store this in a named variable so it is more clear why this should always be 1.0.

cwschilly (approved):
Looks good, just one comment to potentially improve clarity

src/lbaf/Applications/LBAF_app.py
ppebay: Actually, this is a good finding... as the min/max arrangements must be modified to take into account the **local** alphas. I suggest taking this to a separate issue, if you can create one @cwschilly. Thanks again!

Issue #550, opened 2024-09-28
Investigate removing `estimate` method in CriterionBase


cwschilly:
As mentioned in PR #546

`CriterionBase` defines an `estimate` method, but it may not be used.

- `TemperedCriterion` doesn't implement it at all (which causes a pylint warning)
- `StrictLocalizingCriterion` implements it by simply forwarding it to the compute function:

```python
def estimate(self, r_src: Rank, o_src: list, *args) -> float:
        """Estimate is compute because all information is local for this criterion."""
        return self.compute(r_src, o_src, *args)
```

Investigate if we can just remove the `estimate` method entirely.

ppebay:
This is resolved by PR #597

PR #597, opened 2025-03-30
#596: Replace hard-coded 1.0 with actual local alpha values for the computation of max arrangement work statistics

### LINKED ISSUE #596: Replace hard-coded 1.0 with actual local alpha values for the computation of max arrangement work statistics
## Description of the issue:
As reported by @cwschilly, currently in `LBAF_app`, the `compute_min_max_arrangements_work()` is called with all `alpha` values set to 1.

## How to fix the issue:
This must be replaced with the use of the proper rank-local `alpha` values when available

ppebay:
Resolve #596 #550 

**N.B. for reviewers:** upon closer inspection, it does not make much sense to update this method which in fact is now seldom (if ever) used, as the MILP advantageously replaces it by allowing to compute optima in a much more scalable way: the brute force option calling the max arrangement computation does not work past a few ranks and objects due to combinatorial explosion. The code has thus been refactored to entirely do away with this baseline comparison approach which _de facto_ was no longer used -- and causes substantial maintenance challenges.

As a side note I suggest we make MILP callable directly from LBAF (which we discussed in the past @lifflander and for which some very preliminary work had been done but that has not been touched in quite a while now).

### DIFF
--- a/config/conf.yaml
+++ b/config/conf.yaml
@@ -13,7 +13,6 @@ work_model:
     gamma: 0.0
 
 # Specify algorithm
-brute_force_optimization: true
 algorithm:
   name: InformAndTransfer
   phase_id: 0
--- a/config/synthetic-blocks.yaml
+++ b/config/synthetic-blocks.yaml
@@ -16,7 +16,6 @@ work_model:
       max_memory_usage: 45.0
 
 # Specify algorithm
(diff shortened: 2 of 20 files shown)


lifflander (approved):
Looks great to me!

PR #602, opened 2025-05-15
#601: Fix bug in output console

### LINKED ISSUE #601: Fix bug in output console
While verifying results of cluster swapping an error in the reporting of cluster swaps was discovered

ppebay:
Resolves #601


cwschilly (approved):
LGTM

PR #612, opened 2025-07-21
#611: Implement all update formulae and make them optionally usable

### LINKED ISSUE #611: Implement all update formulae and make them optionally usable
This is in order to:
1.  make LBAF have a behavior closer to that of vt
2. verify the formulae of the ACM-ICPP'25 paper
3. make this versatile by allowing to switch between "global information" computation of the transfer criterion (faster but less vt-esque) vs. the "limited information" one (more vt-esque but more complex).

ppebay:
Resolves #611

### DIFF
--- a/config/synthetic-blocks.yaml
+++ b/config/synthetic-blocks.yaml
@@ -14,7 +14,7 @@ work_model:
     gamma: 0.0
     delta: 0.0
     upper_bounds:
-      max_memory_usage: 45.0
+      max_memory_usage: 10.0
 
 # Specify algorithm
 algorithm:
@@ -45,7 +45,7 @@ visualization:
   y_ranks: 2
   z_ranks: 1
   object_jitter: 0.5
-  rank_qoi: homing
+  rank_qoi: max_memory_usage
   object_qoi: shared_id
   save_meshes: true
   force_continuous_object_qoi: true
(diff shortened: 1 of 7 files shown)


ppebay:
@lifflander this is now fully working, and the optional choice to use the update formulae (or not) is determined by the use of the `TemperedWithUpdates` criterion, instead of the full-information `Tempered` one. Eventually, we may simply deprecate the latter and replace it with the former. 

Currently, the debug verification is performed, by comparing the update vs. full recompute results; until we convince ourselves fully that the update formulae are correctly implemented (and no theoretical errors were made in the paper either...).

Follow-on issues as created to keep track of these 2 points.

ppebay:
@lifflander I made some additional testing following our recent conversation, no problem found. Still awaiting your review when you get a chance.

config/synthetic-blocks.yaml
```diff
@@ -10,9 +10,9 @@ check_schema: true
 work_model:
   name: AffineCombination
   parameters:
-    beta: 0.0
+    beta: 1000.0
```
lifflander: Do we want to change this synthetic blocks configuration file?

lifflander (approved):
Overall, this looks great. I've walked through all the update formula and I think they are correct.

config/synthetic-blocks.yaml
ppebay: Thanks for noticing, I pushed back the original configuration. But meanwhile discovered when testing with different values that we have a problem with memory limits. Switching PR to draft as a result. Will re-request review later

PR #616, opened 2025-10-16
#615: spec: add new methods to adding communication edges to spec

### LINKED ISSUE #615: Add comm to json specification
We don't have a good way to add communication edges.

lifflander:
Fixes #615


cwschilly (approved):
All changes look good. There are a fair amount of pylint warnings from the code that was already here. Maybe we should resolve those in a separate issue?

ppebay (approved):

PR #618, opened 2025-10-21
#617: Writing out json is very slow with large files

### LINKED ISSUE #617: Writing out json is very slow with large files
Need to benchmark to find out why it is slow. Might be related to communication.

cwschilly:
Fixes #617 

Explanation: Since writing JSON is a fairly small task, Python multiprocessing was introducing unneeded overhead. It's much faster to just write everything in serial: for the problem in question, the `write()` call went from taking 5 minutes down to 4 seconds

### DIFF
--- a/src/lbaf/IO/lbsVTDataWriter.py
+++ b/src/lbaf/IO/lbsVTDataWriter.py
@@ -96,7 +96,7 @@ class VTDataWriter:
         try:
             self.__extension = parameters["json_output_suffix"]
             self.__compress = parameters["compressed"]
-            self.__add_communications = parameters["communications"]
+            self.__add_communications = parameters.get("communications", True)
         except Exception as e:
             self.__logger.error(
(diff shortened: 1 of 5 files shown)


cwschilly:
I wrote a simple `timer` decorator for profiling that may be useful down the line, so I separated it out into its own file and kept the decorators in VT Data Writer. The timings are only written when the logging level is `debug`.

src/lbaf/IO/lbsVTDataWriter.py
```diff
@@ -94,11 +96,13 @@ def __init__(
         try:
             self.__extension = parameters["json_output_suffix"]
             self.__compress = parameters["compressed"]
+            self.__add_communications = parameters.get("communications", True)
```
cwschilly, line 99: Does it make sense to default to `True` here? Or should users opt-in to writing comms?

PR #619, opened 2026-03-25
#613: Replace Tempered criterion with TemperedWithUpdates implementation and deprecate the latter

### LINKED ISSUE #613: Replace Tempered criterion with TemperedWithUpdates implementation and deprecate the latter
Following #611 and as explained in PR #612 , the `TemperedWithUpdates` criterion, using updates instead of the full-information recomputing approach, should replace the`Tempered` one.

ppebay:
resolves #613


cwschilly (approved):
LGTM

